PR #41, opened 2018-04-25
Update README.md

jay-hodgson:
Remove reference to DynamoDB.  Change to download json data (converted and uploaded to Synapse), and added --jsonArray parameter to mongoimport command since I was getting an error the error:
> Failed: error unmarshaling bytes on document #0: JSON decoder out of sync - data changing underfoot?


Zaknarfen:
Thanks @jay-hodgson !

PR #258, opened 2018-07-12
scripts for getting data

kdaily:
Added some scripts that `npm` can run.

`data:local` assumes you have synapse set up and credentials stored somewhere.
`data:local-aws` gets credentials from AWS secrets but requires that you have an AWS credentials profile called `agora` with AWS secret keys set.
`data:aws` assumes running on AWS EC2 with an IAM role that has access to the secrets (our current EC2 has a role attached to get these permissions).

Happy to take feedback on how to implement this better.

This creates files with different names from the ones that exist in the codebase now:

- `data/rnaseq_differential_expression.json` for the differential expression widget
- `data/network.json` for the network
- `data/team_info.json` for the team info
- `data/gene_info.json` for the gene info


kdaily:
need to add `synapseclient` and `awscli` as dependencies in the README (`pip install synapseclient awscli`)

PR #300, opened 2018-07-18
remove unused files

kdaily:
@Zaknarfen @gatof81 make sure we really aren't using any of these!


Zaknarfen:
I could not find references to those files. Merging.

PR #545, opened 2018-10-31
force agora to use https

zaro0508:
Setup nginx to redirect requests from http to https


xschildw (approved):
Looks good to me from reading the AWS docs.

PR #779, opened 2020-02-07
fix(build): don't rename css

conundrumgirl:


config/webpack.prod.js
```diff
@@ -171,7 +171,7 @@ module.exports = function (env) {
      * See: http://webpack.github.io/docs/configuration.html#plugins
      */
     plugins: [
-      new MiniCssExtractPlugin({ filename: '[name]-[hash].css', chunkFilename: '[name]-[chunkhash].css' }),
+      new MiniCssExtractPlugin({ filename: '[name].css', chunkFilename: '[name]-[chunkhash].css' }),
       new HashedModuleIdsPlugin(),
       new webpack.LoaderOptionsPlugin({
```
zaro0508, line 176: does `postcss` option depend on the MiniCssExtractPlugin?  the one in webpack.dev.js does not contain this option.

conundrumgirl: MiniCssExtractPlugin is only used in prod configuration

zaro0508: sorry if i'm not clear.  my question is whether there is a dependency between options in the  LoaderOptionsPlugin and the miniExtractPlugin.  To be more specific if you changed a config in MiniCssExtractPlugin does that break the autoprefixer thing that's in the LoderOptionsPlugin?

conundrumgirl:
taking different approach

PR #787, opened 2020-02-12
attempt an incremental node version update

zaro0508:


conundrumgirl (approved):
smart move! Keep incrementing until we hit the wall:) Please let me know when we need to bump agora to 9

zaro0508:
I've verified that deploying with nodejs v12.4.1 will work so no need to try this.  Will just update to solution stack "64bit Amazon Linux 2018.03 v4.13.0 running Node.js" running Node.js v12.4.1

PR #790, opened 2020-03-19
fix(AG-19 AG-21): druggability text updates and go.MF rename to go_MF

conundrumgirl:
just some text updates, go.MF column rename and lint fixes


zaro0508 (changes requested):
looks like a bunch of test failures?

conundrumgirl:
yeah.. the tests expected the old text. Fixed now.

zaro0508 (approved):

PR #807, opened 2020-07-09
Ag-65: Add new gene correlation plot

emmanmills:
Screen Shot 2020-07-09 at 12 27 18 PM


src/app/charts/candlestick-chart/candlestick-chart-view/candlestick-chart-view.component.ts
```diff
@@ -0,0 +1,250 @@
...
+            .style('top', (d3.event.pageY + 30) + 'px');
+    }
+
+    hideTooltip(tooltip) {
+        tooltip.transition()
+            .duration(500)
+            .style('opacity', 0);
+    }
+
+    // Helper functions for testing
```
conundrumgirl, line 237: I don't think there should be code in the component with the sole purpose of just facilitating testing. Is there any way to avoid that? I understand that with charts testing can be  difficult  so I understand if it's not

src/server/routes/api.ts
```diff
@@ -121,6 +124,17 @@ connection.once('open', () => {
             g.study = g.study;
             g.tissue = g.tissue;
 
+            // TODO: this map may not be needed
```
conundrumgirl, line 127: What will determine it?

src/server/routes/api.ts
```diff
@@ -1121,6 +1145,23 @@ connection.once('open', () => {
...
     };
+
+    const getGeneCorrelationData = (hgncId: string) => {
+        if (!genesNeuroCorr.length) {  // if DB has error
+            console.log('getGeneCorrelationData: Mongo DB return document length ', genesNeuroCorr.length);
+            return [];
+        }
+
+        const ensembledIds = hgncToEnsembl.get(hgncId);
+        if (ensembledIds !== undefined) {
```
conundrumgirl, line 1157: if (ensembledIds === undefined)  is there an error? Should we output something?

 
           
        }

conundrumgirl (approved):
looks good. A couple of comments but nothing major

src/app/charts/candlestick-chart/candlestick-chart-view/candlestick-chart-view.component.ts
emmanmills: Yes, I can remove the private keyword and my test should be able to run without these helper functions.

src/server/routes/api.ts
emmanmills: Yes, will update or remove the comment once we get the newest data file. Thanks for reminding me! :)

src/server/routes/api.ts
emmanmills: No, but I did forget to return a value if the ensembleIDs is undefined but I have just fixed it now. Thanks for catching that!

PR #874, opened 2021-09-23
Remove unsupported settings

emmanmills:
Remove unsupported settings for EBS node.js server based on what is documented here: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options-specific.html#command-options-nodejs

### DIFF
--- a/.ebextensions/eb-options.config
+++ b/.ebextensions/eb-options.config
@@ -1,8 +1,3 @@
 option_settings:
   aws:elasticbeanstalk:container:nodejs:
     ProxyServer: "nginx"
-    NodeCommand: "node --max_old_space_size=2000  ./dist/server.js"
-    NodeVersion: "12.14.1"
-
-  aws:elasticbeanstalk:command:
-    Timeout: 2700


zaro0508 (approved):

.ebextensions/eb-options.config
```diff
@@ -1,8 +1,3 @@
 option_settings:
   aws:elasticbeanstalk:container:nodejs:
     ProxyServer: "nginx"
-    NodeCommand: "node --max_old_space_size=2000  ./dist/server.js"
-    NodeVersion: "12.14.1"
-
-  aws:elasticbeanstalk:command:
```
zaro0508, line 7: you probably don't want to remove this.

zaro0508 (changes requested):

.ebextensions/eb-options.config
emmanmills: @zaro0508 , which lines you don't want me to remove? The timeout? Thanks.

PR #884, opened 2021-10-08
AG-207:  Add histogram

emmanmills:
The title on top of each chart is for demo only. Should be replaced by the correct title bar. 

Screen Shot 2021-10-08 at 12 34 24 PM
Screen Shot 2021-10-08 at 12 38 56 PM


JessterB:
We need to make one tiny change to the text here since we are dropping the Neuropath score. It should now read:

The TREAT-AD Center at Emory-Sage-SGC has developed a target ranking score encompassing genomics, genetics, and literature evidence.

PR #905, opened 2021-12-20
Remove unused code & clean up

JessterB:
* AG-249: Remove unused routes
* AG-279: Remove mongoose write schemas, use lean queries
* AG-257: Remove references to unused Go_MF data
* AG-231/AG-276: Clean up linting & deprecation warnings

Lean queries: https://mongoosejs.com/docs/tutorials/lean.html


src/server/routes/api.ts
```diff
@@ -172,47 +159,35 @@ connection.once('open', () => {
...
             }
         });
-        await allTissues.sort();
-        await allModels.sort();
 
         //////////////////////////////////////////////////////////////////////////////////////////
     });
 
-    GenesInfo.find({ nominations: { $gt: 0 } })
+    GenesInfo.find({ nominations: { $gt: 0 } }).lean()
```
nickgros, line 185: Not super familar with `lean` (or mongoose in general) but seems safe to me since these services are all getters and there are no setters.

Shouldn't break anything if we're not using custom getters (I have no idea if we are). Wondering how well TypeScript handles this change and if it catches any possible errors introduced by this change

src/app/schemas/gene.ts
```diff
@@ -1,67 +1,7 @@
...
-// Handler **must** take 3 parameters: the error that occurred, the document
-// in question, and the `next()` function
-GeneSchema.post('save', function(error, doc, next) {
-    if (error.name === 'MongoError' && error.code === 11000) {
-        next(new Error('There was a duplicate key error'));
-    } else {
-        next(error);
-    }
-});
+export let GeneSchema: Schema = new Schema({}, { collection: 'genes' });
```
nickgros, line 4: I'm guessing the above code was unnecessary + duplicated from models?

nickgros: Ditto for all the other similar changes in the schema folder

src/server/routes/api.ts
```diff
@@ -172,47 +159,35 @@ connection.once('open', () => {
...
         if (err) {
             next(err);
         } else {
             tableGenesById = genes.slice();
-            // Table genes ordered by nominations
-            tableGenesByNom = await genes.slice().sort((a, b) => {
-                return (a.nominations > b.nominations) ? 1 :
-                    ((b.nominations > a.nominations) ? -1 : 0);
-            });
-
```
nickgros: Assuming this sort isn't needed anymore

src/server/routes/api.ts
```diff
@@ -620,7 +558,7 @@ connection.once('open', () => {
     // Get all gene infos that match an id, using the hgnc_symbol or ensembl id
     router.get('/gene/infos/:id', (req, res, next) => {
         const fieldName = (req.params.id.startsWith('ENSG')) ? 'ensembl_gene_id' : 'hgnc_symbol';
-        const isEnsembl = (req.params.id.startsWith('ENSG')) ? true : false;
```
nickgros, line 645: ...amazing

src/server/routes/api.ts
```diff
@@ -732,35 +670,38 @@ connection.once('open', () => {
...
-                                const flinks = [...links, ...linkB, ...linksC];
-                                res.setHeader(
-                                    'Cache-Control', 'no-cache, no-store, must-revalidate'
-                                );
-                                res.setHeader('Pragma', 'no-cache');
-                                res.setHeader('Expires', 0);
-                                res.json({ items: flinks });
-                            }
+
+                    GenesLinks.find({geneB_ensembl_gene_id: req.params.id}, async (errB, linkB) => {
```
nickgros, line 706: @JessterB I understand that you just changed formatting/added `lean` but if you understand this, can you put a comment? It's not clear to me why we're making up to 3 DB calls as I'd only expect 1 here

nickgros (approved):
Had some questions above but I have no objections

src/server/routes/api.ts
JessterB: I haven't didn't see any custom getters in the code, and the branch looks fine when I run locally...I'm not sure about the TypeScript question.

src/app/schemas/gene.ts
JessterB: It turns out that mongoose schemas are only required to specify write behaviors (https://stackoverflow.com/questions/28630865/mongoose-read-only-without-schema), and since we don't actually ever write to the database all of that code was unnecessary and confusing since some of the schemas were out of sync with the actual data in the db.

src/server/routes/api.ts
JessterB: lol

src/server/routes/api.ts
JessterB: `tableGenesByNom` wasn't used anywhere, so I just removed it entirely.

PR #909, opened 2021-12-22
Uncouple experimental validation teams from nominating teams

JessterB:


src/app/genes/gene-overview/exp-validation/exp-validation.component.ts
```diff
@@ -26,12 +27,12 @@ export class ExpValidationComponent implements OnInit {
     }
 
     LoadTeam() {
-        const teams = this.geneService.getCurrentTeams();
-        if (teams.length) {
-            this.teamInfo = this.data.map((item, i) => {
-               return teams.filter(j => j.team === item.team)[0];
+       this.apiService.getAllTeams()
```
nickgros: Do we have to `getAll` here or can we query `getTeams`?

JessterB: The `getTeams` method only looks at nominatedtarget.teams, which is the root cause the bug -- when viewing genes that haven't been nominated, the previously viewed gene's nominatedtarget.teams is still cached, but may not contain the teams we need. It would probably be better to load and cache all of the teams on startup, then use that cache for both experimental validation and nominations display.

JessterB: Ok, I think this is a nicer fix - apiService.getTeams now takes a list of strings, and each component calls it with the appropriately derived list of team names. I also added a Teams cache!

nickgros (approved):

PR #910, opened 2022-01-06
Genes without genes documents now load from the nominated targets list

JessterB:
* Add additional GeneService test cases
* Replace deprecated TestBed.get() with TestBed.inject()  -> See https://angular.io/api/core/testing/TestBed#get

### DIFF
--- a/src/app/core/services/gene.service.ts
+++ b/src/app/core/services/gene.service.ts
@@ -77,8 +77,8 @@ export class GeneService {
         this.setCurrentGene(data.item);
         this.setCurrentInfo(data.info);
         this.setCurrentExpValidation(data.expValidation);
-        this.setCurrentModel(data.item.model);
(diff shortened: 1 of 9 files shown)


src/app/core/services/gene.service.ts
```diff
@@ -77,8 +77,8 @@ export class GeneService {
         this.setCurrentGene(data.item);
         this.setCurrentInfo(data.info);
         this.setCurrentExpValidation(data.expValidation);
-        this.setCurrentModel(data.item.model);
-        this.setCurrentTissue(data.item.tissue);
+        this.setCurrentModel(data.item?.model);
+        this.setCurrentTissue(data.item?.tissue);
         this.setCurrentGeneOverallScores(data.overallScores);
     }
```
JessterB, line 84: This 2-character change is the bug fix

nickgros (approved):

PR #911, opened 2022-01-07
Use isXInADBrain booleans

JessterB:
* AG-263: Display new booleans in  gene network sidepanel & similar genes table
* AG-290: Replace all instances of legacy isChangedInADBrain with new isAnyRNAChangedInADBrain


src/app/core/services/gene.service.spec.ts
```diff
@@ -43,16 +43,17 @@ describe('Service: Gene: TestBed', () => {
...
+    // it('should update the previous gene with the current gene', () => {
+    //     const upgSpy = spyOn(geneService, 'updatePreviousGene').and.callThrough();
+    //     geneService.setCurrentGene(mockGene1);
+    //     expect(geneService.getCurrentGene()).toEqual(mockGene1);
+    //     expect(geneService.getPreviousGene()).toEqual(undefined);
+    //
+    //     geneService.updatePreviousGene();
+    //     expect(upgSpy).toHaveBeenCalled();
+    //     expect(geneService.getPreviousGene()).toEqual(mockGene1);
+    // });
```
mfazza: Why not just delete this?

src/app/genes/gene-similar/gene-similar.component.ts
```diff
@@ -302,8 +305,8 @@ export class GeneSimilarComponent implements OnInit {
                     return this.druggability(rowObj.druggability);
                 }
             } else {
-                if (field === 'isIGAP' || field === 'isChangedInADBrain' ||
-                    field === 'haseqtl') {
+                if (field === 'isIGAP' || field === 'isAnyRNAChangedInADBrain' ||
+                    field === 'isAnyPRoteinChangedInADBrain' || field === 'haseqtl') {
```
mfazza: Oh, that's what Anna was talking about yesterday

mfazza (approved):
💯 🥇

src/app/core/services/gene.service.spec.ts
JessterB: I can move the test case into the related JIRA ticket if having it in here makes you sad!

JessterB:
Found a bug, putting this into draft state while I sort it out...

src/app/core/services/gene.service.spec.ts
JessterB: I used the built-in 'xit' feature to suppress the test from execution, so this is better now.

PR #912, opened 2022-01-08
Nominated targets table displays default value when no druggabilty.ph…

JessterB:
…aros_class

### DIFF
--- a/src/app/genes/genes-list/genes-list.component.ts
+++ b/src/app/genes/genes-list/genes-list.component.ts
@@ -117,17 +117,20 @@ export class GenesListComponent implements OnInit {
                     .sort((a: string, b: string) => a.localeCompare(b)).join(', ') : '';
                 de.input_data = (inputDataArray.length) ? inputDataArray.filter(this.getUnique)
                     .sort((a: string, b: string) => a.localeCompare(b)).join(', ') : '';
-                de.validation_study_details = (validationStudyDetailsArray.length) ? [...new Set(validationStudyDetailsArray)]
+                de.validation_study_details = (validationStudyDetailsArray.length) ?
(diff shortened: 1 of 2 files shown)


src/app/genes/genes-list/genes-list.component.ts
```diff
@@ -117,25 +117,27 @@ export class GenesListComponent implements OnInit {
...
 
                 // Druggability columns
+                const noValue = 'No value';
                 if (de.druggability && de.druggability.length) {
                     de.sm_druggability_bucket =
                         de.druggability[0].sm_druggability_bucket.toString();
                     de.safety_bucket = de.druggability[0].safety_bucket.toString();
                     de.abability_bucket = de.druggability[0].abability_bucket.toString();
-                    de.pharos_class = de.druggability[0].pharos_class;
+                    de.pharos_class = de.druggability[0].pharos_class ? de.druggability[0].pharos_class : noValue;
```
nickgros, line 133: Seems like `de.pharos_class` should probably be something like `de.displayValue`, but I'm missing a lot of context so I'll leave that up to you.

nickgros (approved):

src/app/genes/genes-list/genes-list.component.ts
JessterB: The pharos class is a value defined by NIH (https://pharos.nih.gov/targets); renaming it would likely just introduce confusion.

JessterB: Oooh now I merged I understand what you mean here, luckily (??) I have to do more fixing in here to get table sorting working properly so I'll address this feedback in that upcoming PR.

PR #920, opened 2022-01-18
AG-291: Display Resources tab when no druggability data

JessterB:
* AG-259: Always display the Resources tab
* Display reasonable score defaults when no druggability data

### DIFF
--- a/src/app/genes/gene-overview/gene-druggability/gene-druggability.component.html
+++ b/src/app/genes/gene-overview/gene-druggability/gene-druggability.component.html
@@ -20,7 +20,7 @@
                     </div>
                     <div class="col-md-3 col-sm-4 col-4 text-center header-score">
                         <div [ngStyle]="getBucketIconStyle(false)" [style.background-color]="getBucketBGColor(geneInfo.druggability[0].sm_druggability_bucket, 'sm')" [style.color]="getBucketTextColor(geneInfo.druggability[0].sm_druggability_bucket, 'sm')">
-                            <div class="icon-description">{{geneInfo.druggability[0].sm_druggability_bucket}}</div>
+                            <div class="icon-description">{{getBucketNumberText(geneInfo.druggability[0].sm_druggability_bucket)}}</div>
                         </div>
                     </div>
                     <div class="col-md-6 col-sm-12 col-12 header-description">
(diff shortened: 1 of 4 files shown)


src/app/genes/gene-overview/gene-druggability/gene-druggability.component.html
```diff
@@ -20,7 +20,7 @@ <h2>
                     </div>
                     <div class="col-md-3 col-sm-4 col-4 text-center header-score">
                         <div [ngStyle]="getBucketIconStyle(false)" [style.background-color]="getBucketBGColor(geneInfo.druggability[0].sm_druggability_bucket, 'sm')" [style.color]="getBucketTextColor(geneInfo.druggability[0].sm_druggability_bucket, 'sm')">
-                            <div class="icon-description">{{geneInfo.druggability[0].sm_druggability_bucket}}</div>
+                            <div class="icon-description">{{geneInfo.druggability[0].sm_druggability_bucket !== 100 ? geneInfo.druggability[0].sm_druggability_bucket : '&ndash;'}}</div>
```
jay-hodgson: Recommend keeping this logic in the typescript file rather than the html.  Perhaps a helper function, like "getFriendlyBucketValue (and pass sm_druggability_bucket, abability_bucket, and safety_bucket to it).
Also in the typescript, you could define this special constant value that you're using, DEFAULT_BUCKET_VALUE=100

src/app/genes/gene-overview/gene-druggability/gene-druggability.component.ts
```diff
@@ -36,6 +36,9 @@ export class GeneDruggabilityComponent implements OnInit {
     classes: string[] = ['A', 'B', 'C', 'D', 'E', 'F'];
     additionalResources: GeneResourceType[] = [];
 
+    noDataTitle = 'Not analyzed';
```
jay-hodgson: These could also be constants.  Looks like they are defined as class variables, but I don't think I've seen this kind of declaration before (without the var/let/const keyword).

jay-hodgson (changes requested):
Few minor changes.  No logical issues, just an effort to keep the code clean and values properly scoped.

src/app/genes/gene-overview/gene-druggability/gene-druggability.component.html
JessterB: That makes sense, done and done

src/app/genes/gene-overview/gene-druggability/gene-druggability.component.ts
JessterB: That seems to be a feature of typescript class members (see e.g. https://stackoverflow.com/questions/37265275/how-to-implement-class-constants).  I switched to using get methods to avoid the entire issue here!

PR #924, opened 2022-01-20
AG-282 - Hidden row graph tick for negative values

mattcgenui:


src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
```diff
@@ -355,14 +355,18 @@ export class RowChartViewComponent implements OnInit, OnDestroy, AfterViewInit,
     updateXTicks(chart: dc.RowChart) {
         const allTicks = chart.selectAll('g.axis g.tick');
         allTicks.each(function(t, i) {
+            // Handle UTF-8 characters
+            let value = parseFloat(d3.select(this).select('text').text());
+            value = isNaN(value) ? parseFloat('-' + d3.select(this).select('text').text().replace(/[^,.0-9]/g, '')) : value;
```
jay-hodgson, line 360: Please add a comment whenever there's a regular expression involved. Something as simple as
"if text is not a number, then replace all non-numeric values with the empty string" would suffice in this case.

jay-hodgson:
Thank you for creating a new clean branch from develop for each issue as well!

jay-hodgson (approved):

PR #925, opened 2022-01-20
AG-256 - Changed how height is calculated for related genes "hairball" graph

mattcgenui:
Instead of setting the height depending on the right column, the height is calculated depending on the number of nodes with a minimum of 400. This should make the "hairball" centered in all cases.


jay-hodgson (approved):
Agree that Jess should test this approach to the height calculation, thanks Matt!

PR #928, opened 2022-01-24
Draft - Node update

mattcgenui:


jay-hodgson:
I don't really understand how the infra is created for Agora.  This should update node on travis, and local.

jay-hodgson (approved):

PR #930, opened 2022-01-24
AG-304 - Updated node to 14.18.2

mattcgenui:


JessterB:
Let's hold off on merging this until we are ready for the Elastic Beanstalk platform upgrade: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.platform.upgrade.html#using-features.platform.upgrade.bluegreen

JessterB:
we aren't going to a beanstalk platform version that supports node 14 this time

PR #932, opened 2022-01-25
AG-51 - Added link (back to gene) to similar genes page

mattcgenui:


src/app/genes/gene-similar/gene-similar.component.ts
```diff
@@ -135,6 +135,7 @@ export class GeneSimilarComponent implements OnInit {
                     this.geneService.updateGeneData(data);
                     this.gene = data.item;
                     this.geneInfo = data.info;
+                    console.log(this.gene);
```
jay-hodgson: Was this for testing purposes?

mattcgenui: Yes, thank you for catching that!

jay-hodgson (approved):

PR #935, opened 2022-01-25
AG-218 - Changed terms and privacy link (Footer/Mobile navbar)

mattcgenui:


src/app/core/footer/footer.component.scss
```diff
@@ -130,8 +130,14 @@
     font-size: $font-size-regular;
     color: $show-more-color;
     cursor: pointer;
+    text-decoration: none;
```
jay-hodgson: I wonder if this justifies a new mixin to be included here (and anywhere links should not be underlined until hover).  Does not block merge (approving this PR).

jay-hodgson (approved):

src/app/core/footer/footer.component.scss
mattcgenui: Good idea, I added a mixin.

PR #938, opened 2022-01-26
AG-246 - Implemented optimizations for gene search

mattcgenui:
- Added 2 characters minimum requirement
- Increased debounce time from 300ms to 500ms
- Added validation for Ensembl Gene ID (to prevent 502 response)

### DIFF
--- a/src/app/genes/gene-search/gene-search.component.ts
+++ b/src/app/genes/gene-search/gene-search.component.ts
@@ -25,13 +25,17 @@ export class GeneSearchComponent implements OnInit {
     @Input() style: any;
     @Input() infos: Gene[];
     queryField: FormControl = new FormControl();
+    currentQuery: string = '';
+    isSearching: boolean = false;
+    isEnsemblIdSearch: boolean = false;
+    isNotFound: boolean = false;
     results: GeneInfo[] = [];
     hasFocus: boolean = false;
-    isSearching: boolean = false;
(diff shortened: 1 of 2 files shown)


src/app/genes/gene-search/gene-search.component.ts
```diff
@@ -46,20 +50,11 @@ export class GeneSearchComponent implements OnInit {
...
                 distinctUntilChanged(),
                 switchMap((query) => {
                     this.results = [];
-
-                    if (query) {
-                        this.currentQuery = query;
-                        this.isSearching = true;
-                        return this.search(query);
-                    } else {
-                        this.currentQuery = '';
```
jay-hodgson, line 59: Why is this no longer needed?  Are there cases when query is not set and a reset is necessary?

jay-hodgson: Tagging @mattcgenui

mattcgenui: Sorry @jay-hodgson, I missed this feedback. This code was move to line 114/115: 
https://github.com/Sage-Bionetworks/Agora/pull/938/files#diff-1304225f209da309cb2ce390fc0e905f8af131af9b36f957ea3dbc53b2369e38R114

jay-hodgson (approved):

PR #947, opened 2022-02-09
Precalculated distribution data

mattcgenui:


src/app/charts/box-plot/box-plot-view/box-plot-view.component.ts
```diff
@@ -220,10 +209,10 @@ export class BoxPlotViewComponent implements OnInit, OnDestroy, AfterViewInit {
         });
     }
 
-    getChartPromise(): Promise<dc.BoxPlot> {
+    getChartPromise(): Promise<any> {
```
jay-hodgson: Not typed?

src/app/charts/box-plot/box-plot-view/box-plot-view.component.ts
```diff
@@ -163,31 +168,15 @@ export class BoxPlotViewComponent implements OnInit, OnDestroy, AfterViewInit {
 
         const bpGroup = {
             all() {
-                const currentGenes = self.dataService.getGeneEntries().slice().filter((g) => {
-                    return g.model === self.geneService.getCurrentModel();
+                const distributionData = self.dataService.getRnaDistributionData().filter((data) => {
```
jay-hodgson, line 170: So much better

src/app/core/services/api.service.ts
```diff
@@ -244,4 +245,15 @@ export class ApiService {
 
         return this.http.get(baseUrl, { headers, params });
     }
+
+    getRnaDistributionData(): Observable<RnaDistribution[]> {
+        const headers = new HttpHeaders({
```
jay-hodgson: This could be a constant that is used across the app (I'd guess that the same headers are used for the majority of the service calls).  Named something like JSON_RESPONSE_NO_CACHE_HEADERS

src/app/core/services/data.service.ts
```diff
@@ -27,6 +27,7 @@ export class DataService {
     modelsDim: any;
     dbgenes: Observable<Gene[]>;
     geneEntries: Gene[];
+    rnaDistributionData: any;
```
jay-hodgson: Missing type

jay-hodgson (approved):

src/app/core/services/api.service.ts
mattcgenui: Added `const` on top of the file, I don't think it is used anywhere else

PR #955, opened 2022-02-11
Added new route for evidence data and updated RNA plots

mattcgenui:


jay-hodgson (approved):

jay-hodgson:
I see the new service definition and use

PR #957, opened 2022-02-11
Remove legacy aws:elasticbeanstalk:container:nodejs NodeCommand

JessterB:


.ebextensions/eb-options.config
```diff
@@ -1,7 +1,6 @@
 option_settings:
   aws:elasticbeanstalk:container:nodejs:
     ProxyServer: "nginx"
-    NodeCommand: "node --max_old_space_size=2000  ./dist/server.js"
```
zaro0508, line 4: umm, i think you still need this command?  i believe the docs indicate that you just need to move it to package.json https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/nodejs-platform-dependencies.html

zaro0508 (changes requested):

.ebextensions/eb-options.config
JessterB: The project's package.json file already had the command: https://github.com/Sage-Bionetworks/Agora/blob/develop/package.json

zaro0508: looks good then.

zaro0508 (approved):

PR #960, opened 2022-02-18
AG-332 - Added negative values support for median expression plot

mattcgenui:


JessterB (commented):
One small display issue I noticed here is that we have "-0" on the y-axis, which isn't a thing.  If we don't shift the axis instead of inverting (see next comment), we would ideally display "0" without a negative sign.

JessterB:
Is shifting the y-axis up so the bars can project downwards a possibility? Inverting the axis makes it appear that these values are significant, when they are the opposite. We'll also need to remove the red significance line from the inverted plot if we continue to present the data upside down. If we can draw the negative bars downwards, the red line can stay where it is and will clearly show that the negative values are well below the threshold.

PR #961, opened 2022-02-18
AG-335 - Added ENSG id on typeahead lookups

mattcgenui:


JessterB:
With these changes the app still always loads the first ENSG, but I think that will be fixed when we merge https://github.com/Sage-Bionetworks/Agora/pull/963

jay-hodgson:
Looks great.  Happy to see you changed it to show the errorMessage as well!
Interestingly, this counts map is very similar to a phone screen coding question that we typically ask :)

jay-hodgson (approved):

PR #963, opened 2022-02-22
AG-229 / AG-374 - Use ensembl_gene_id instead of hgnc_symbol in routes and code

mattcgenui:


src/app/genes/genes-list/genes-list.component.ts
```diff
@@ -231,20 +231,20 @@ export class GenesListComponent implements OnInit {
         this.msgs = [{
             severity: 'info',
             summary: 'Gene Selected',
-            detail: 'Gene: ' + event.data.hgnc_symbol
+            detail: 'Gene: ' + (event.data.hgnc_symbol || event.data.ensembl_gene_id)
         }];
```
JessterB, line 235: We also need to display the ENSG in the table's Gene Symbol column instead of the hgnc; the column is blank when a nominated gene has no hgnc value.

See around line 45 of this file, which I can't put a comment on because GitHub.

conundrumgirl (approved):
We are removing  volcano-plot?

PR #966, opened 2022-02-24
Load teamInfo values by team name rather than by expVal report index

JessterB:
Loading teamInfo values by expVal index fails when a gene has multiple expVal reports submitted by the same team.

### DIFF
--- a/src/app/genes/gene-overview/exp-validation/exp-validation.component.html
+++ b/src/app/genes/gene-overview/exp-validation/exp-validation.component.html
@@ -12,8 +12,12 @@
             <h1 class="inner-section-divider"></h1>
         </div>
         <div class="section-thin no-padding-top overview-sections">
-            <h2>{{getTeamInfoField(d.team, "program")}} : {{getTeamInfoField(d.team, "team_full")}}</h2>
-            <p class="col-8 section-desc">{{getTeamInfoField(d.team, "description")}}</p>
+            <div *ngFor="let t of teamInfo;">
+                <div *ngIf="t.team === d.team ">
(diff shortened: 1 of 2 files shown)


src/app/genes/gene-overview/exp-validation/exp-validation.component.html
```diff
@@ -12,8 +12,8 @@ <h2 *ngIf="data">Experimental Validation of {{d.hgnc_symbol}}</h2>
             <h1 class="inner-section-divider"></h1>
         </div>
         <div class="section-thin no-padding-top overview-sections">
-            <h2>{{teamInfo[i].program}}: {{teamInfo[i].team_full}}</h2>
-            <p class="col-8 section-desc">{{teamInfo[i].description}}</p>
+            <h2>{{getTeamInfoField(d.team, "program")}} : {{getTeamInfoField(d.team, "team_full")}}</h2>
```
jay-hodgson: Each of these getTeamInfoField calls iterates over all items in teamInfo, so I'm sure there's a more efficient way to do this using Angular patterns.  Maybe wrap with something like:
```
<div *ngFor="let t of teamInfo;">
  <div *ngIf="t.name === d.team ">
    <h2>{{t.program}} : {{t.team_full}}</h2>
...

```

jay-hodgson: This would iterate over teamInfo items once (instead three times) for every item in data.

JessterB: That does seem like a much better approach, thanks for the suggestion!

PR #980, opened 2022-03-14
AG-418/AG-419 - Fixed refresh/update issues on consistency of change plot

mattcgenui:


src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
```diff
@@ -235,12 +235,10 @@ export class RowChartViewComponent implements OnInit, OnDestroy, AfterViewInit,
                     }
                 })
                 .on('renderlet', function(chart) {
-                    if (self.firstRender) {
-                        self.firstRender = false;
-                        // Copy all vertical texts to another div, so they don't get hidden by
-                        // the row chart svg after being translated
-                        self.moveTextToElement(chart, self.stdCol.nativeElement, 9);
-                    }
+                    self.firstRender = false;
```
JessterB: I pulled down this branch and ran it locally, and I noticed that there is something going on with the x-axis label. See my comment on AG-419 for details. I'm not certain that it's due to the change you made here around self-firstRender, it's just a guess...

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
```diff
@@ -235,12 +235,10 @@ export class RowChartViewComponent implements OnInit, OnDestroy, AfterViewInit,
                     }
                 })
                 .on('renderlet', function(chart) {
-                    if (self.firstRender) {
```
jay-hodgson, line 238: firstRender is initialized to true, and set to false here, but is it actually used/referenced anywhere (after this change)?

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
```diff
@@ -302,22 +308,18 @@ export class RowChartViewComponent implements OnInit, OnDestroy, AfterViewInit,
         const squareSize = 18;
         const lineWidth = 60;
 
-        // Test if we should display the chart. Using this variable so we don't see
-        // the rows rectangles change into small squares abruptly
-        if (!self.display) {
```
jay-hodgson, line 307: Same question for the display variable.

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
```diff
@@ -400,28 +403,29 @@ export class RowChartViewComponent implements OnInit, OnDestroy, AfterViewInit,
...
-            textGroup.append(function() {
-                return n;
-            });
+        const container = d3.select(el).html('');
+        const svg = container.append('svg');
+        const group = svg.append('g').attr('class', 'textGroup');
+        const texts: any = chart.selectAll('g.row > text');
+
+        texts.each(function() {
+            this.style.display = 'none';
```
jay-hodgson, line 404: Looks like it used to remove these nodes from the DOM, and now it hides them.  I think this is fine for this case, but just something to note.

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
mattcgenui: Thank you, I removed the unused code

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
mattcgenui: Thank you, I removed the unused code

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
mattcgenui: Thank you, I removed the unused code

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
mattcgenui: Yes I believe I had to do that otherwise the text values are missing on refresh

jay-hodgson (approved):

PR #984, opened 2022-03-16
AG-428/AG-340 - Sort and columns changes on nominated and similar gene tables

mattcgenui:


JessterB:
I noticed some visual issues while running this branch locally, see AG-428 for screen shots

JessterB:
I noticed some visual issues while running this branch locally, see AG-428 for screen shots

src/app/genes/gene-similar/gene-similar.component.ts
```diff
@@ -241,6 +241,11 @@ export class GeneSimilarComponent implements OnInit {
     }
 
     reorderValues(event) {
+        const defaultCol = this.selectedColumns.find(d => d.field === 'hgnc_symbol');
+        if (!defaultCol) {
+            this.selectedColumns.unshift(this.cols[0]);
```
jay-hodgson: What's going on here?  You're inserting this.cols[0] to the selectedColumns if you don't find hgnc_symbol?  What is in this.cols[0]?

src/app/genes/gene-similar/gene-similar.component.ts
```diff
@@ -371,8 +376,16 @@ export class GeneSimilarComponent implements OnInit {
     customSort(event: SortEvent) {
         event.data.sort((data1, data2) => {
             let result = null;
-            const value1 = data1[event.field];
-            const value2 = data2[event.field];
+            let value1 = null;
+            let value2 = null;
+
+            if ('hgnc_symbol' === event.field) {
+                value1 = data1.hgnc_symbol || data1.ensembl_gene_id;
```
jay-hodgson: If hgnc_symbol really is null or undefined when not set, I recommend using the nullish coalescing operator (??) instead of OR (||) when implementing a default value.  The empty string (""), 0, and false are also [falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)), and in some contexts these are valid values.

mattcgenui: Good point!

src/app/genes/gene-similar/gene-similar.component.ts
mattcgenui: I made to code more detailed, is it meant to keep the name column always selected/displayed

jay-hodgson (approved):

PR #985, opened 2022-03-16
AG-343/AG-366 - Color and layout changes on gene network plot

mattcgenui:


JessterB:
I pulled this down and ran it locally, and the colors are now updating properly! 

However, the sidebar content still doesn't update properly when you reselect the 'main' gene (but the link does go to the 'main' gene's results).

JessterB:
I pulled this down and ran it locally, and the colors are now updating properly! 

However, the sidebar content still doesn't update properly when you reselect the 'main' gene (but the link does go to the 'main' gene's results).

src/app/genes/gene-overview/evidence-menu/rna/gene-network/gene-network.component.ts
```diff
@@ -156,30 +156,28 @@ export class GeneNetworkComponent implements OnInit {
 
     updategene(event: Gene) {
         const self = this;
-        if (event.ensembl_gene_id !== this.geneService.getCurrentGene().ensembl_gene_id) {
```
jay-hodgson, line 159: Did this used to ignore a no-op?

mattcgenui: Yes, if you click a different gene in the network and click back on the current gene, the sidebar data does not update. There is probably room for improvement by caching the data and avoid reloading at every click

jay-hodgson (approved):

PR #988, opened 2022-03-25
AG-285 - Added global error handling

mattcgenui:


conundrumgirl (approved):

src/app/shared/services/error-handler.service.ts
```diff
@@ -0,0 +1,25 @@
...
+    constructor(
+        private injector: Injector,
+        private messageService: MessageService
+    ) { }
+
+    handleError(error) {
+        const self = this;
+
+        this.messageService.clear();
+        this.messageService.add({
```
jay-hodgson, line 15: This is great, but will not help us track down the error in the wild.  I'd like to see the error also sent to the console (json stringified if appropriate) to give us more information if we're unable to reproduce the issue.

PR #1002, opened 2022-04-06
Gene comparison tool fixes and changes batch 2

mattcgenui:


src/app/core/services/gene.service.ts
```diff
@@ -165,6 +165,43 @@ export class GeneService {
         // return this.geneTissues;
     }
 
+    getGeneTissueName(tissue: string): string {
```
JessterB: A few nitpicks here:
* Some of these aren't tissues (BRAAK, CERAD, and the 2 Diagnoses), getTooltipText() might be a better name for the method?
* AntPFC and MFG aren't surfaced in the app, I'm just curious where you found those as I don't see any corresponding moved/removed code in this PR.

JessterB (approved):

PR #1009, opened 2022-04-12
AG-490 - Moved toast to bottom

mattcgenui:


src/app/app.component.ts
```diff
@@ -49,6 +51,16 @@ export class AppComponent implements OnInit {
                 event: 'pageView',
                 url: e.urlAfterRedirects
             });
+
+            const genesRouter = e.urlAfterRedirects.match(/\(genes-router:([^)]+)\)/);
+            const pageClass = 'page-' + (
+                genesRouter && genesRouter.length > 1 && genesRouter[1]
+                ? genesRouter[1].split('/')[0]
```
jay-hodgson: Does this need to be this complicated, to ultimately result in the css class page-gene-comparison-tool?

jay-hodgson (approved):

PR #1010, opened 2022-04-12
Log errors to the console

JessterB:


src/app/shared/services/error-handler.service.ts
```diff
@@ -20,6 +20,7 @@ export class ErrorHandlerService implements ErrorHandler {
         });
         setTimeout(() => { self.messageService.clear(); }, 5000);
 
+        console.error(error);
```
jay-hodgson: I've seen too many cases of this resulting in the output "[object Object]".  So I'd recommend changing this to:
console.error(JSON.stringify(error))

JessterB: Done!

jay-hodgson (approved):

PR #1015, opened 2022-04-15
AG-509: Handle missing proteomics data correctly

JessterB:


src/app/genes/gene-overview/evidence-menu/proteomics/proteomics.component.ts
```diff
@@ -40,7 +40,7 @@ export class ProteomicsComponent implements OnInit {
     ngOnInit() {
         this.gene = this.geneService.getCurrentGene();
         this.geneInfo = this.geneService.getCurrentInfo();
-        if (this.gene && this.geneService.getGeneProteomics()) { this.isEmptyGene = false; }
```
jay-hodgson, line 43: Used to check that this.geneService.getGeneProteomics() is defined.  Is it possible that this is undefined?  (If so, this call to length could fail)

JessterB: I don't think so? The field is initialized with an empty array:

`    geneProteomics: Proteomics[] = [];`

And the setter takes an array:

```
    setGeneProteomics(proteomics: Proteomics[]) {
        this.geneProteomics = proteomics;
    }
```

...but I put that check back in anyway, just in case.

jay-hodgson (approved):

PR #1019, opened 2022-04-18
AG-515: Gene search error messages wrap and do not highlight

JessterB:
See AG-515 for the before and after screen shots. It looks much better to me with these changes, but I'm far from a CSS wizard...

### DIFF
--- a/src/app/genes/gene-search/gene-search.component.html
+++ b/src/app/genes/gene-search/gene-search.component.html
@@ -26,11 +26,11 @@
         <ul *ngIf="errorMessage"  class="filter-select">
             <li class="filter-select-error" [class.italic-result]="isNotFound">
                 <div class="gene-name row">
-                    <span class="gene-search-error">
-                        <div class="col-md-12 col-sm-12 col-12">
+                    <div class="col-md-12 col-sm-12 col-12">
+                        <span class="gene-search-error">
                             {{errorMessage}}
-                        </div>
-                    </span>
+                        </span>
+                    </div>
                 </div>
             </li>
         </ul>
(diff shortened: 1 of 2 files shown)


src/app/genes/gene-search/gene-search.component.html
```diff
@@ -24,11 +24,13 @@
             </div>
         </div>
         <ul *ngIf="errorMessage"  class="filter-select">
-            <li class="filter-select-list" [class.italic-result]="isNotFound">
+            <li class="filter-select-error" [class.italic-result]="isNotFound">
                 <div class="gene-name row">
-                    <div class="col-md-12 col-sm-12 col-12">
-                        {{errorMessage}}
-                    </div>
+                    <span class="gene-search-error">
```
jay-hodgson: Bit dangerous putting this between the bootstrap grid definitions (row and col-*).  Does it still work if you put span inside the col div (a child of the <div class="col-md-12 col-sm-12 col-12" /> instead)?

src/app/genes/gene-search/gene-search.component.scss
```diff
@@ -117,14 +117,15 @@
     margin-right: 30px;
 }
 
+.filter-select-error,
 .filter-select-list {
     cursor: pointer;
     padding: 5px 10px;
-    height: 60px;
+    height: auto;
```
jay-hodgson: What's the height if you do not set it to auto here?  Is it not auto?

jay-hodgson (approved):

src/app/genes/gene-search/gene-search.component.scss
JessterB: It is! Good catch, I've removed the unnecessary height attribute.

src/app/genes/gene-search/gene-search.component.html
JessterB: I moved the span inside the div, and tweaked a few things to get it working.

PR #1022, opened 2022-04-22
RNA/proteomics box plot tooltips pull identifiers from geneInfo

JessterB:


src/app/charts/box-plot/pbox-plot-view/pbox-plot-view.component.ts
```diff
@@ -135,7 +135,7 @@ export class PBoxPlotViewComponent implements OnInit, OnDestroy, AfterViewInit {
 
     getGeneSymbol(): string {
         const geneSymbol = this.geneService.getCurrentInfo().hgnc_symbol
-            ?? this.geneService.getCurrentInfo().ensembl_gene_id;
+            || this.geneService.getCurrentInfo().ensembl_gene_id;
         return geneSymbol;
```
JessterB, line 139: Note that the nullish coalescing operator didn't work for us here, as we want to fall back to ENSG when hgnc is a blank string.

jay-hodgson (approved):

PR #1024, opened 2022-05-07
AG-473: Use geneInfo as source of truth for hgnc

JessterB:


src/app/genes/gene-similar/gene-similar.component.ts
```diff
@@ -204,13 +204,6 @@ export class GeneSimilarComponent implements OnInit {
...
                     de.nominated_target_display_value = de.nominations > 0;
 
-                    // Populate MedianExpression display fields
-                    const meArray = (de.medianexpression.length) ?
-                        de.medianexpression.map((me: MedianExpression) => me.tissue) : [];
-                    de.brain_regions_display_value = (meArray.length) ? meArray.filter(this.getUnique)
-                        .sort((a: string, b: string) => a.localeCompare(b)).join(', ') : '';
-                    de.num_brain_regions_display_value = meArray.length.toString();
-
                     // Populate Druggability display fields
```
JessterB, line 207: These are no longer used

PR #1033, opened 2022-05-24
AG-511 - Reenable download component

JessterB:
This looks terrible and some tests are failing...

### DIFF
--- a/src/app/shared/components/download/download.component.scss
+++ b/src/app/shared/components/download/download.component.scss
@@ -69,15 +69,15 @@
         transition: $transition-duration;
 
         .ui-radiobutton-icon {
-            display: flex !important;
+            display: flex;
             width: 100%;
             height: 100%;
             align-items: center;
             justify-content: center;
             transition: $transition-duration;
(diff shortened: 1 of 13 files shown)


src/app/shared/components/download/download.component.scss
```diff
@@ -56,44 +56,59 @@
...
+        transition: $transition-duration;
+
+        .ui-radiobutton-icon {
+            display: flex !important;
+            width: 100%;
+            height: 100%;
+            align-items: center;
+            justify-content: center;
+            transition: $transition-duration;
+            top: 0 !important;
```
jay-hodgson: Heads up, using !important is widely discouraged.  It can be used in desperation, but in almost all cases we can devise a css selector whose specificity will cause it to take effect.

jay-hodgson (approved):

src/app/shared/components/download/download.component.scss
mattcgenui: :+1:

PR #1038, opened 2022-05-25
AG 435 - Fixed inconsistent logfc rounding

mattcgenui:


src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
```diff
@@ -191,8 +191,7 @@ export class RowChartViewComponent implements OnInit, OnDestroy, AfterViewInit,
             const chartInst = dc.rowChart(self.rowChart.nativeElement)
                 .gap(4)
                 .title(function(d) {
-                    return 'Log Fold Change: ' +
-                        self.dataService.getSignificantFigures(+d.value.logfc);
+                    return 'Log Fold Change: ' + d.value.logfc;
```
jay-hodgson, line 194: Is getSignificantFigures() still used anywhere (is there more code to delete)?

jay-hodgson (approved):

src/app/charts/row-chart/row-chart-view/row-chart-view.component.ts
JessterB: It is still used in a few other places, so unfortunately we can't delete it.

PR #1039, opened 2022-05-25
AG 544 - Scale box plot y axis

mattcgenui:


src/app/charts/box-plot/box-plot-view/box-plot-view.component.ts
```diff
@@ -184,12 +184,19 @@ export class BoxPlotViewComponent implements OnInit, OnDestroy, AfterViewInit {
                         return d.tissue === gene.tissue;
                     });
 
-                    if (data) {
-                        data['key'] = data['tissue'];
-                        data['value'] = [data['min'], data['median'], data['max']];
-                    }
-
-                    return data;
+                    return {
```
jay-hodgson: Possible that data is undefined (previously there was a check)?  Setting up the data object LGTM (including axis min/max logic)

src/app/shared/services/plot.service.ts
```diff
@@ -516,13 +516,13 @@ export class PlotHelperService {
 
         _chart.yAxisMin = function() {
             const padding = _yRangePadding * yAxisRangeRatio();
-            const min = parseInt(d3.min(_chart.data(), function(g) { return g['min']; }), 10) - 1;
+            const min = parseInt(d3.min(_chart.data(), function(g) { return g['axis_min'] || g['min']; }), 10) - 1;
```
jay-hodgson: Is gene.logfc not always defined?  When is g['axis_min'] or g['axis_max'] undefined, requiring this fallback value?

mattcgenui: This file contains an "extended/tweaked" version of D3 Boxplot which can be used to generate any type of box plot. In this case, the plot has an additional layer with dots (gene.logfc) generated independently from the boxes min/max. Using "axis_min" and "axis_max" gives options to override the y scale only if needed. Does that make sense?

jay-hodgson (approved):

PR #1044, opened 2022-05-26
AG 470 - Remove gene symbol from column options

mattcgenui:


src/app/genes/gene-similar/gene-similar.component.ts
```diff
@@ -57,6 +57,7 @@ export class GeneSimilarComponent implements OnInit {
         { field: 'safety_rating_display_value', header: 'Safety Rating' },
         { field: 'ab_modality_display_value', header: 'Antibody Modality' },
     ];
+    optionalCols: any[] = this.cols.filter(col => col.field !== 'hgnc_symbol');
```
jay-hodgson: Like to see this refactored so that both cases use the same function (if additional columns need to be filtered out in the future, the developer is likely to miss one instance).

jay-hodgson (approved):

PR #1056, opened 2022-06-03
AG-578: eQTL links go to the published paper rather than the preprint

JessterB:


JessterB:
The updated link was validated locally, so I'm self-approving this content-only PR.

PR #1087, opened 2022-08-25
AG-811: Fix issue with mobile nav on initial load

JessterB:
https://sagebionetworks.jira.com/browse/AG-811


src/app/features/genes/components/gene-details/gene-details.component.ts
```diff
@@ -200,6 +199,10 @@ export class GeneDetailsComponent implements OnInit, AfterViewInit {
     }, 100);
   }
 
+  ngAfterViewChecked() {
```
jay-hodgson, line 202: That's another useful angular lifecycle hook

jay-hodgson (approved):

PR #1089, opened 2022-08-29
AG-802: Close hamburger menu when menu item selected

JessterB:
https://sagebionetworks.jira.com/browse/AG-802

### DIFF
--- a/src/app/core/components/header/header.component.html
+++ b/src/app/core/components/header/header.component.html
@@ -96,7 +96,7 @@
       </button>
       <div class="header-nav-inner">
         <ul>
-          <li *ngFor="let item of navItems">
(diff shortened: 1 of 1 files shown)


jay-hodgson (approved):
Out of curiosity, would this also work if you set a single click handler on the ul element instead?

JessterB:
Yes, yes it does. Good call, I've updated the PR.

JessterB:
After testing that change a bit more, I realized that it prevents you from using the search box in the hamburger menu, since the menu closes when you click on it. I've reverted the change and we will stick with the two separate click handlers.
I also discovered that the menu would close if you clicked on whitespace rather than on a navigation link, so I've moved the new click handler down to the anchor level to prevent that.

PR #1091, opened 2022-09-01
AG-802: Close mobile nav on search navigation

JessterB:


src/app/core/components/header/header.component.ts
```diff
@@ -70,4 +70,8 @@ export class HeaderComponent implements OnInit {
   toggleNav() {
     this.isShown = !this.isShown;
   }
+
+  onSearchNavigation() {
```
jay-hodgson: Why is this new function needed, rather than just calling toggleNav()?

JessterB: Excellent question. It isn't needed, so I removed it.

jay-hodgson (approved):

PR #1094, opened 2022-09-08
AG-831: Update y-axis break handling to work with unsorted distributions

JessterB:
The previous logic only compared each value to the next value. This logic worked fine when our values in the underlying data were accidentally sorted in descending order, but once the data was corrected to return the values in the correct order for their associated bins, that logic broke.

### DIFF
--- a/src/app/features/charts/components/score-chart/score-chart.component.ts
+++ b/src/app/features/charts/components/score-chart/score-chart.component.ts
@@ -88,8 +88,11 @@ export class ScoreChartComponent extends BaseChartComponent {
     });
   }
 
-  getMinDiff(count: number, distribution: any[]) {
-    const min = Math.min(...distribution.map(d => count - d.value).filter(v => v > 0));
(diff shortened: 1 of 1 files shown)


src/app/features/charts/components/score-chart/score-chart.component.ts
```diff
@@ -91,6 +88,11 @@ export class ScoreChartComponent extends BaseChartComponent {
     });
   }
 
+  getMinDiff(count: number, distribution: any[]) {
+    const min = Math.min(...distribution.map(d => count - d.value).filter(v => v > 0));
```
jay-hodgson: This is a really dense line to read.  Please consider breaking this logic into multiple lines.

JessterB: I've updated the method and added some documentation to make it easier to follow

jay-hodgson (approved):

PR #1098, opened 2022-09-12
AG-835: Bump up timeout duration for tracking beanstalk deployment status

JessterB:
This should help avoid false fails in travis that occurr when deployments taking longer than the default timeout duration of 10 min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ jobs:
         - npm run ci:clean
         - ./scripts/set-agora-version.sh || travis_terminate 1
         - zip /tmp/agora-app.zip -qr9 * .[^.]*
-        - eb deploy --profile prod --verbose --region ${AWS_REGION} --staged agora-staging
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         - npm run ci:clean
         - ./scripts/set-agora-version.sh || travis_terminate 1
         - zip /tmp/agora-app.zip -qr9 * .[^.]*
-        - eb deploy --profile prod --verbose --region ${AWS_REGION} --staged agora-staging
+        - eb deploy --profile prod --verbose --region ${AWS_REGION} --staged agora-staging --timeout 15
```
zaro0508: i think `--timeout` needs to go before the `agora-staging` eb profile.  

Also you might want to apply this to prod deployment as well because [prod is set to deploy to more instances](https://github.com/Sage-Bionetworks/agora2-infra/blob/main/config/agoraprod/prod/agora.yaml#L14).

zaro0508 (changes requested):

xschildw (approved):

PR #1102, opened 2022-10-28
AG-707 Adding ability to clear search boxes

sagely1:


src/app/features/genes/components/gene-search/gene-search.component.scss
```diff
@@ -69,6 +69,29 @@
...
+    .gene-search-bar-close {
+      button {
+        @include reset-button;
+        
+        position: absolute;
+        top: 0;
+        right: 0;
+        width: 40px;
+        height: 100%;
+        cursor: pointer;
```
conundrumgirl, line 81: do we need to specify a cursor on a button? Is it different from default behavior?

conundrumgirl:
I am not familiar with the code so it might be a silly question, but can we do without the showSearchResults boolean and instead just set the result to [] or the unfiltered record when the search is cleared? Logically it shouldn't mean show/hide rather than filter results/don't filter.

src/app/features/genes/components/gene-search/gene-search.component.scss
sagely1: Good question.  Yes, this will enable this button to have the standard cursor as the other buttons on the page.  Without this line, the default seems to be no change on mouseover.

conundrumgirl: Makes sense. So the others have pointer and button default is the arrow so you're just making them match. 
 If it is the desired behavior we want universally, does it make sense to have a global button: hover { cursor: pointer} so that we don't have to do it for every button?

sagely1: Absolutely!  That's a great idea.  I was thinking it might make sense to do some global styles cleanup as part of a future cleanup since this is probably one of many repeated styles and improvements I would like to make as I get familiar with the code.  My changes only affect this component so the nice thing is we would only have to test this component for this PR.  I'll create a separate issue in the backlog for the repeated styles.

conundrumgirl (approved):

PR #1118, opened 2022-11-28
first pass at security vulnerabilities

sagely1:


sagely1:
Addressed a few vulnerabilities but more can potentially be addressed in the future (ideally after moving to Angular 15).

JessterB (approved):

PR #1123, opened 2022-11-30
AG-908 Ensembl paste bug in search box

sagely1:


src/app/features/genes/components/gene-search/gene-search.component.ts
```diff
@@ -147,6 +147,7 @@ export class GeneSearchComponent implements AfterViewInit {
     this.input.nativeElement.blur();
     this.query = '';
     this.results = [];
+    this.showGeneResults = false;
```
JessterB, line 150: This looks like the same change that was in the other PR (AG-906). Did you push up the wrong branch maybe?

JessterB (changes requested):

PR #1130, opened 2022-12-20
AG-901 removed Literature score bar chart and updated copy

sagely1:


src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.ts
```diff
@@ -42,12 +42,15 @@ export class GeneSoeChartsComponent implements OnInit {
...
     this.geneService.getDistribution().subscribe((data: any) => {
-      const overallScoreDistribution = data.overall_scores;
+      let overallScoreDistribution = data.overall_scores;
 
       overallScoreDistribution.sort((a: any, b: any) =>
         a.name > b.name ? 1 : -1
       );
 
+      //remove literature score
+      overallScoreDistribution = overallScoreDistribution.filter((item: any) => (item.name !== 'Literature Score'));
```
sagely1, line 52: MongoDB has this score as part of the API call to get gene scores.  At this point in time, it is expected that Literature Score will return, so this line is to filter out Literature Score until a future time when it will be reimplemented.

conundrumgirl (approved):

PR #1131, opened 2022-12-21
AG-911 Updated tooltip copy

sagely1:


src/app/features/genes/components/gene-nominated-targets/gene-nominated-targets.component.html
```diff
@@ -76,7 +76,7 @@ <h1 class="h2">Nominated Target List</h1>
           [columns]="columns"
           [searchTerm]="searchTerm"
           [gctLink]="{ nominations: nominations.join(',') }"
-          gctLinkTooltip="Use Agora's Gene Comparison Tool to compare all nominated genes in this list."
+          gctLinkTooltip="Use Agora's Gene Comparison Tool to compare all nominated genes in this list"
```
JessterB: This should also be reworded as a call to action: "Compare all nominated genes in this list using Agora’s Gene Comparison Tool"

JessterB (changes requested):

src/app/features/genes/components/gene-nominated-targets/gene-nominated-targets.component.html
sagely1: Fixed missing copy

JessterB (approved):

JessterB:
I think this one is ok to merge to dev and release with 3.1.1 since it's just a bug fix.

PR #1138, opened 2023-01-10
[IT-2407] redirect http to https

zaro0508:
Setup the http to https redirect in node[1]

[1] https://www.geeksforgeeks.org/how-to-detect-http-or-https-then-force-redirect-to-https-in-javascript/

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -111,6 +111,14 @@ if (app.get('env') === 'development') {
   });
 }
 
+//http to https redirects
+app.get('/', (req: Request, res: Response) => {
+    if (req.protocol == 'http') {
+        res.redirect('https://' +
+        req.get('host') + req.originalUrl);
+    }
+});
+
 // -------------------------------------------------------------------------- //
 // Server
 // -------------------------------------------------------------------------- //
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -113,10 +113,10 @@ if (app.get('env') === 'development') {
 
 //http to https redirects
 app.get('/', (req: Request, res: Response) => {
-    if (req.protocol == 'http') {
-        res.redirect('https://' +
-        req.get('host') + req.originalUrl);
-    }
+  if (req.protocol == 'http') {
+    res.redirect('https://' +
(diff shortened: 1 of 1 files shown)


src/server/server.ts
```diff
@@ -30,11 +30,9 @@ app.set('trust proxy', 'loopback');
 
 //http to https redirects
 app.enable('trust proxy');
-app.use((req,res,next)=>{
-  if(rec.secure){
-    next();
-  }else{
-    res.redirect('https://'+req.host+req.url);
+app.use(function(request, response, next)=>{
+  if (request.protcol == 'http') {
```
xschildw: protocol

Q: can't we just add listener to the ALB?

zaro0508: yes, however AWS does not make it easy to automate and that's why i have decided to try to do it in node.

xschildw (approved):

JessterB:
Unfortunately I don't think that this change will address our root issue, which is that the 'Share URL' button does not work when this page is loaded over http: http://agora.adknowledgeportal.org/genes/comparison

It fails because we can't access `navigator.clipboard` over http: https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined

I believe that this PR would only redirect requests to the server (/api/\*), however we need the web app's URLs (/* pages) to be https to access `navigator.clipboard` on the page.

Note that the app already makes https requests to the server, even when the app itself is running over http:

Screen Shot 2023-01-11 at 11 03 53 AM

The stackoverflow thread I linked above also includes a workaround using an off-screen text area, but that solution requires using a deprecated command (https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) that has been replaced by the `navigator.clipboard` API...

JessterB:
This might be worth doing anyway if we want to prevent users from hitting our server API directly over http, although I don't know if anyone does that, or if we care about insecure access to our read-only APIs that return only open public data.

JessterB (approved):
Let's see what this does in dev, since validating it on localhost would require a fair amount of additional effort.

PR #1145, opened 2023-01-26
AG-948 nominations ordering bug

sagely1:
Moved nominations sorting logic into api and optimized ordering on the frontend.

### DIFF
--- a/src/app/features/teams/components/team-list/team-list.component.ts
+++ b/src/app/features/teams/components/team-list/team-list.component.ts
@@ -22,20 +22,7 @@ export class TeamListComponent {
     return this._teams;
   }
   @Input() set teams(teams: Team[]) {
-    this.sort(teams);
     this._teams = teams;
-  }
-
-  sort(teams: Team[]) {
-    teams.sort((a, b) => {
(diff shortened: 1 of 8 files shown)


src/app/features/teams/components/team-list/team-list.component.ts
```diff
@@ -22,22 +22,9 @@ export class TeamListComponent {
     return this._teams;
   }
   @Input() set teams(teams: Team[]) {
-    this.sort(teams);
```
sagely1, line 25: Removing sort here as this will be sorted from the api

src/server/components/teams.ts
```diff
@@ -32,11 +32,30 @@ export async function getTeams() {
   }
 
   teams = await TeamCollection.find().lean().exec();
+  //sort null programs last
+  sortNullProgramsForTeamsLast(teams);
```
sagely1, line 36: api will now sort nulls last (default behavior is to sort nulls first).  This way, Sage will always be last as Sage is (by design) the only entry with a null program.

conundrumgirl (approved):

PR #1155, opened 2023-02-13
AG-920 add numeric scores columns

sagely1:


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -223,6 +238,17 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
     columns.sort();
     this.columns = columns;
 
+    // add scores columns
```
conundrumgirl, line 250: nitpick but may be more elegant
```
for (const name  of  ['GENETIC','MULTI-OMICS','RISK SCORE']) {
 if (!columns.includes(name))
      columns.unshift(name);
}
```

src/server/components/scores.ts
```diff
@@ -0,0 +1,85 @@
...
+// Internal
+// -------------------------------------------------------------------------- //
+import { cache } from '../helpers';
+import { Scores, ScoresCollection } from '../models';
+
+// -------------------------------------------------------------------------- //
+// Functions
+// -------------------------------------------------------------------------- //
+
+export async function getAllScores() {
```
conundrumgirl, line 11: it seems like these three functions are very similar. Can we combine them into a parametrized single function

conundrumgirl (commented):
A couple of comments

src/server/components/scores.ts
sagely1: whoops.  This should've been deleted since we don't need use any of these methods.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: I'll refactor this.  Looks like I was using it in two places

sagely1 (commented):
Deleted extraneous methods, added missing tests and DRY'd up GCT column name variable.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -222,7 +236,21 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
 
     columns.sort();
     this.columns = columns;
-
+    // add scores columns
+    this.COLUMN_NAMES_WITH_NUMERICS.forEach((columnName) => {
+      if (!columns.includes(columnName))
+        columns.unshift(columnName);  
+    });
+    const geneticColumnName = 'GENETIC';
```
conundrumgirl: shouldn't this be removed?

conundrumgirl (commented):
i think there is some extra code there

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: Good catch.  I think I removed it and hit undo.

conundrumgirl (approved):

PR #1163, opened 2023-03-28
AG-983 updated alignment and responsiveness of site version and data version

sagely1:
Footer design has been updated.  Site version and data version are now aligned with other footer links and take up less vertical space.  Elements will still wrap as the window size shrinks.


hallieswan (approved):
Looks good to me!

PR #1165, opened 2023-04-04
AG-985 make bar chart color configurable

sagely1:


src/app/features/charts/components/score-chart/score-chart.component.ts
```diff
@@ -30,10 +30,12 @@ export class ScoreChartComponent extends BaseChartComponent {
     this.init();
   }
 
+  @Input() barColor = '#8b8ad1';
```
sagely1, line 33: Going forward, more CSS props will be passed in.  In this case, we are using this component for each score chart, so we have a default color, but a color can be passed in to override the default color.

src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.ts
```diff
@@ -73,6 +77,13 @@ export class GeneSoeChartsComponent implements OnInit {
     });
   }
 
+  getBarColor(title: string) {
+    if (title === 'Genetic Risk Score') {
```
sagely1: The color is determined by whether the score chart is the "Genetic Risk Score".  Green in this case but otherwise lavender.

src/app/features/charts/components/score-chart/score-chart.component.ts
```diff
@@ -129,7 +131,7 @@ export class ScoreChartComponent extends BaseChartComponent {
       );
 
     // Colors
-    this.chart.colors(['#8b8ad1']);
+    this.chart.colors(this.barColor);
```
nickgros: Type seems to have changed from `string[]` to `string`, not sure if that matters. I don't see this property used anywhere.

nickgros (approved):

src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.ts
JessterB: The Target Risk Score should be the one with the different color, not the Genetic Risk Score

JessterB (changes requested):

src/app/features/charts/components/score-chart/score-chart.component.ts
sagely1: Yeah, the official docs mention an array but it didn't seem to affect the chart to use a string.  I will switch this to an array just to be safe.

src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.ts
sagely1: Fixed!

JessterB (approved):
Looks good to me!

PR #1171, opened 2023-04-14
AG-978 zero scores showing as data unavailable

sagely1:


src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.ts
```diff
@@ -105,13 +105,13 @@ export class GeneSoeChartsComponent implements OnInit {
       this._gene?.overall_scores || ({} as OverallScores);
 
     if ('Genetic Risk Score' === name) {
-      return scores['genetics_score'] || 0;
```
sagely1, line 108: This is the underlying issue it seems.  Null scores have meaning to the charts.

src/app/features/charts/components/score-chart/score-chart.component.spec.ts
```diff
@@ -39,7 +39,7 @@ describe('Component: Chart - Score', () => {
   });
 
   it('should display message if not data', () => {
-    expect(component.score).toEqual(0);
+    expect(component.score).toEqual(null);
     expect(element.querySelector('.chart-no-data')).toBeTruthy();
   });
```
nickgros: You could consider also verifying that a 0 score is treated as an actual value, if you want to capture both sides of the edge case. Might be kind of awkward with how this test is set up.

src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.scss
```diff
@@ -52,12 +52,12 @@
...
-    padding-bottom: 15px;
-  }
+  // .score-chart {
+  //   min-width: 320px;
+  // }
+
+  // .chart-body {
+  //   padding-top: 15px;
+  //   padding-bottom: 15px;
+  // }
```
nickgros: Delete if this won't be used in the near-future 🙂

nickgros (approved):
Few suggestions but otherwise LGTM!

src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.scss
sagely1: Might as well delete this since I think this whole chart will be rewritten.

src/app/features/charts/components/score-chart/score-chart.component.spec.ts
sagely1: Great suggestion.  I added zero and valid value cases.

PR #1173, opened 2023-04-26
AG-1051 Create Bio-domains d3 chart

sagely1:


sagely1:
Please also see AG-990 for more requirements

src/app/core/services/helper.service.ts
```diff
@@ -178,4 +178,8 @@ export class HelperService {
     const urlParams = new URLSearchParams(queryString);
     return urlParams.get(name);
   }
+
+  capitalizeFirstLetterOfString(s: string) {
+    return s[0].toUpperCase() + s.slice(1);
```
nickgros, line 185: This will error on the empty string case

src/app/features/charts/components/biodomains-chart/biodomains-chart.component.ts
```diff
@@ -0,0 +1,203 @@
...
+      if (this.selectedIndex > -1)
+        this.selectedBioDomain = this.data[this.selectedIndex].biodomain;
+
+      const svg = d3.select('#chart')
+        .append('svg')
+        .attr('width', 500)
+        .attr('height', 560);
+
+      const chartWidth = +svg.attr('width');
+      const chartHeight = +svg.attr('height');
```
nickgros: Why `=+`? Are these different than the `500` and `560` above?

src/app/features/charts/components/biodomains-chart/biodomains-chart.component.ts
```diff
@@ -0,0 +1,203 @@
...
+          d3.select('#bar-label').select(`text:nth-child(${ index + 1 })`).style('font-weight', 'bold');
+        })
+        .on('mouseout', (event) => {
+          const index = d3.select('#bars').selectAll('rect').nodes().indexOf(event.currentTarget);
+          // if the target is the selected index
+          if (this.selectedIndex !== index) {
+            d3.select('#bars').select(`rect:nth-child(${ index + 1 })`).style('fill-opacity', '50%');
+            d3.select('#bar-label').select(`text:nth-child(${ index + 1 })`).style('font-weight', 'normal');
+          }
+        });
```
nickgros: Looks like it's mostly the same as negative space handlers, I think it might reduce unnecessary duplicate code make sense to just have functions that create each handler 

```ts
function getOnClickHandler(id: "#negative-bars" | "#bars") {
	// ...
}
```

src/app/features/genes/components/gene-biodomains/gene-biodomains.component.scss
```diff
@@ -0,0 +1,65 @@
...
+
+/* Firefox */
+@supports (scrollbar-color: #878E95 transparent) {
+    * {
+        scrollbar-color: #878E95 transparent;
+        scrollbar-width: thin;
+    }
+}
+
+// Custom scrollbar END
```
nickgros: Consider creating and importing a custom sass mixin, instead of just scoping to this component

src/app/features/genes/components/gene-biodomains/gene-biodomains.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    expect(component).toBeTruthy();
+  });
+
+  it('should return an empty string when getGeneName() is called with undefined gene', () => {
+    component.gene = undefined;
+    const result = component.getGeneName();
+    expect(result).toBe('');
+  });
+
+  it('should return gene name when getGeneName() is called with undefined gene', () => {
```
nickgros:
```suggestion
  it('should return gene name when getGeneName() is called with defined gene', () => {
```

src/app/core/services/helper.service.ts
sagely1: Nice catch.  I will check in a few unit tests to exercise the scenario where this a string of length 1 and a string of length 0.

src/app/features/charts/components/biodomains-chart/biodomains-chart.component.ts
sagely1: The + in front of the svg is shorthand for Number() which will convert the string from attr() to a number.  Now that you mention it, I feel like it might just be better to define a const number in the createChart() method since we don't use chartWidth or chartHeight outside of that method.  Also, it should save me from creating an svg element with numeric props only to then have to use attr() to get the string representation and have to convert it back to a number.

src/app/features/genes/components/gene-biodomains/gene-biodomains.component.spec.ts
sagely1: nice catch

src/app/features/charts/components/biodomains-chart/biodomains-chart.component.ts
sagely1: Great suggestion.  I think there might be multiple event handlers I can DRY up. Jess mentioned we may also want to change the hover color of the label text on the bar and negative-bar mouseover so I'll incorporate that feedback with yours.

src/app/features/genes/components/gene-biodomains/gene-biodomains.component.scss
sagely1: I agree. My plan is to rewrite the other charts to remove the dependence on dc.js and revisit this after more details on the Visualization Framework get ironed out.

nickgros (approved):
LGTM!

PR #1174, opened 2023-05-02
AG-986 add biodomains to gene hero

sagely1:


sagely1:
@hallieswan you can just review the commits that start with: "AG-986"

src/app/features/genes/components/gene-hero/gene-hero.component.ts
```diff
@@ -70,4 +71,15 @@ export class GeneHeroComponent {
     }
     return '';
   }
+
+  getBiodomains(): string {
+    if (!this.gene || !this.gene.bio_domains)
+      return '';
+    console.log(this.gene.bio_domains.gene_biodomains);
```
hallieswan: Is the console log message still helpful or could this line be removed?

src/app/features/genes/components/gene-hero/gene-hero.component.ts
```diff
@@ -70,4 +71,15 @@ export class GeneHeroComponent {
...
+    const biodomains = this.gene.bio_domains.gene_biodomains
+      .filter(b => b.pct_linking_terms > 0)
```
hallieswan, line 79: Is this a standard filter to apply to the `gene_biodomains`? Just wondering since I didn't see the filter explicitly called out in the ticket

hallieswan (approved):
Looks good! One question -- is the `biodomain` field guaranteed to be in the "Initial Caps" format requested in the ticket or could it have a different capitalization scheme (e.g. `biodomain: "structural stabilization",`)? If not, could consider adding code to ensure the correct capitalization is applied. I defer to you though, since I'm not familiar with the source data!

src/app/features/genes/components/gene-hero/gene-hero.component.ts
sagely1: You're right, it can be removed.  Good catch!

src/app/features/genes/components/gene-hero/gene-hero.component.ts
sagely1: Good question! There are 19 biodomains as I recall.  Since we always show all of them regardless in the charts, the non-zero ones should be shown in the gene hero.  I'll ask Jess to add that to the description.

PR #1175, opened 2023-05-03
AG-983 fixing footer issues, tooltip box showing in upper left corner

sagely1:


src/app/app.component.scss
```diff
@@ -0,0 +1,9 @@
+#container {
+    display: flex;
+    flex-direction: column;
+    min-height: 100vh;
```
sagely1, line 4: This is to ensure that the footer is always at the bottom of the screen

src/app/core/components/footer/footer.component.scss
```diff
@@ -11,7 +11,7 @@
   justify-content: space-between;
   background-color: var(--color-primary);
   font-weight: 300;
-  padding: var(--spacing-sm);
+  padding: var(--spacing-md);
```
sagely1, line 15: creative shows more padding around elements so I've bumped it to md which is a predefined value (not quite the same as in the creative but close).

src/app/features/charts/components/biodomains-chart/biodomains-chart.component.scss
```diff
@@ -57,14 +57,4 @@ $tooltip-color: #63676C;
...
-
-ul {
-    list-style-type: none;
-    margin: 0;
-    padding: 0;
-
-    li {
-        padding: 0 0 15px;
-    }
 }
```
sagely1, line 60: unused css so cleaning up

src/app/features/genes/components/gene-biodomains/gene-biodomains.component.scss
```diff
@@ -29,12 +40,6 @@ h3 {
     font-weight: bold;
 }
 
-ul {
-    li {
-        font-size: 14px;
-    }
-}
-
```
sagely1, line 43: This was not scoped correctly so the code has been moved to the right scope

src/styles/components/_chart.scss
```diff
@@ -125,7 +125,7 @@
   position: absolute;
   top: 0;
   pointer-events: none;
-  z-index: 999;
+  z-index: 998;
 }
 
 .chart-internal-tooltip {
```
sagely1, line 131: by default, the tooltips are rendered in the top left corner of the screen and the z-index should be set to less than the header to avoid showing up over the header

sagely1, line 131: Ideally, this would be visibly hidden but there are a lot of charts currently in need of a rewrite.  For now, the z-index needs to be set a value less than the header (which is currently 999).

hallieswan (approved):
Looks good!

PR #1181, opened 2023-05-24
Ag 1059 minor adjustments omnibus

sagely1:
@hallieswan You can just review the AG-1059 changes


src/app/features/genes/components/gene-hero/gene-hero.component.html
```diff
@@ -16,18 +16,20 @@ <h1 class="h2 gene-hero-heading">
...
-        </p>
-      </div>
+      <ng-container *ngIf="gene.bio_domains || getAlias()">
+        <hr />
+        <div *ngIf="gene.bio_domains" class="gene-hero-biodomains">
+          <h4>BIOLOGICAL DOMAINS</h4>
+          {{ getBiodomains() }}
+        </div>
+        <div *ngIf="getAlias()" class="gene-hero-aliases">
+          <div class="gene-hero-aliases-heading">Also known as</div>
```
hallieswan, line 26: The ticket says that the 'Biological Domains' header should be the same font size as the 'Also Known As' header -- if the header styles are the same otherwise, could consider also using the `h4` tag for the 'Also Know As' header and removing the `gene-hero-aliases-heading` class, so that the header styling would stay aligned going forward. Not sure if the `gene-hero-aliases-heading` class is used for another purpose though, so up to you!

hallieswan (approved):
AG-1059 changes look good to me!

src/app/features/genes/components/gene-hero/gene-hero.component.html
sagely1: Great suggestion!  I think I will do a pass at css styles in the future.  It would have semantic meaning to have anything labeled header / heading actually use an html header tag.

PR #1185, opened 2023-06-01
AG-1126 gene details aka should be h4

sagely1:
Hi @hallieswan, you can just focus on the AG-1126 commit for the code review.


hallieswan (approved):
AG-1126 looks good to me!

PR #1190, opened 2023-06-08
AG-1125 fixed null score issue

sagely1:


src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -45,26 +45,32 @@ export class ScoreBarChartComponent implements AfterViewInit, OnChanges, OnDestr
...
+    if (
+      (changes.data && !changes.data.firstChange) ||
+      (changes.score && !changes.score.firstChange) ||
+      (changes.barColor && !changes.barColor.firstChange)
+    ) {
+      if (this.score === null) {
+        this.clearChart();
+        this.hideChart();
+      } else {
+        this.clearChart();
```
sagely1, line 58: we need to clear the chart even if the score is not null as the score may be another non-null score and hence we should redraw the chart

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -45,26 +45,32 @@ export class ScoreBarChartComponent implements AfterViewInit, OnChanges, OnDestr
...
-  }
-
   ngAfterViewInit(): void {
-    if (!this.chartRef.nativeElement) {
-      return;
-    }
-
-    if (!this.data)
-      return;
+    if (this.score === null)
```
sagely1, line 66: if the score is null, we need to hide it on initial load, otherwise we create the chart.

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -83,9 +89,6 @@ export class ScoreBarChartComponent implements AfterViewInit, OnChanges, OnDestr
   }
 
   initData() {
-    if (!this.data)
```
sagely1, line 86: duplicate code being removed

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -45,26 +45,32 @@ export class ScoreBarChartComponent implements AfterViewInit, OnChanges, OnDestr
   constructor(private helperService: HelperService) {
   }
 
-  ngOnChanges(): void {
-    if (this.initialized) {
-      // if chart has already been initialized, this means the data has changed
-      this.createChart();
+  ngOnChanges(changes: SimpleChanges): void {
+    if (
```
sagely1, line 49: we need to unfortunately check firstChange === false because on initial load, we don't want to draw the chart twice.  Also, ngOnChanges fires on initial load and subsequent loads

nickgros (approved):

PR #1192, opened 2023-06-08
AG-1141 fixed sort function

sagely1:


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -199,10 +199,10 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
       return gene.target_risk_score;
     }
     if (columnName === this.scoresColumns[1].field) {
-      return gene.multi_omics_score;
+      return gene.genetics_score;
     }
```
hallieswan, line 203: Could the score field names be pulled into constants, then used both in the `scoresColumns` declaration and in `getScoreForNumericColumn`? Just thinking that using a constant (vs an index) would make `getScoreForNumericColumn` more self-documenting and could prevent an issue where changing the ordering of the `scoresColumns` declaration changes this method's behavior. I'm not very familiar with Angular though, so up to you!

hallieswan (approved):
Looks good!

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: I think something along those lines would be great!  However, for the moment I think it might be best to revisit this in the future since I want to change the collection schema in MongoDB.  Some of the columns are for tissues and some of them are for scores.  Since the scores are not alphabetically sorted and have been recently renamed, they have fallen out of sync with the collection naming.  It would be nice to use new properties in a custom class for scores columns that would make sorting them more elegant.

hallieswan: Got it! That sounds like a great solution

PR #1194, opened 2023-06-13
AG-1133 border on left pane wasn't visible

sagely1:


sagely1:
@hallieswan Border on left hand pane wasn't visible.  Was missing border-style.

PR #1204, opened 2023-06-26
AG-1134: score barcharts should take advantage of extra width

hallieswan:
https://github.com/Sage-Bionetworks/Agora/assets/26949006/9759eeb9-8e71-4f7d-9bc9-adb6eaee62ac

### DIFF
--- a/src/app/features/charts/components/score-barchart/score-barchart.component.html
+++ b/src/app/features/charts/components/score-barchart/score-barchart.component.html
@@ -1,7 +1,7 @@
-<div id="score-barchart">
+<div id="score-barchart" #div (window:resize)="onResize(div.getBoundingClientRect().width)">
   <div id="score-barchart-tooltip" class="arrow-below tooltip-arrow" #tooltip (mouseout)="hideTooltip()"></div>
   <svg #chart></svg>
   <div class="chart-no-data" *ngIf="_score === null">
-      <div class="chart-no-data-text">No data is currently available.</div>
+    <div class="chart-no-data-text">No data is currently available.</div>
   </div>
 </div>
--- a/src/app/features/charts/components/score-barchart/score-barchart.component.html
+++ b/src/app/features/charts/components/score-barchart/score-barchart.component.html
@@ -1,4 +1,4 @@
-<div id="score-barchart" #div (window:resize)="onResize(div.getBoundingClientRect().width)">
+<div id="score-barchart" #scoreBarChartContainer>
   <div id="score-barchart-tooltip" class="arrow-below tooltip-arrow" #tooltip (mouseout)="hideTooltip()"></div>
   <svg #chart></svg>
   <div class="chart-no-data" *ngIf="_score === null">
--- a/src/app/features/charts/components/score-barchart/score-barchart.component.scss
+++ b/src/app/features/charts/components/score-barchart/score-barchart.component.scss
@@ -4,6 +4,10 @@ $tooltip-color: #63676C;
     position: relative;
     height: 350px;
 
+    svg {
+        width: 100%; 
+    }
+
     #score-barchart-tooltip {
         position: absolute;
         text-align: center;
--- a/src/app/features/charts/components/score-barchart/score-barchart.component.ts
+++ b/src/app/features/charts/components/score-barchart/score-barchart.component.ts
@@ -34,11 +34,13 @@ export class ScoreBarChartComponent implements OnChanges, AfterViewInit, OnDestr
   @Input() xAxisLabel = 'Gene score';
   @Input() yAxisLabel = 'Number of genes';
 
+  @ViewChild('scoreBarChartContainer') scoreBarChartContainer: ElementRef<HTMLElement> = {} as ElementRef;
   @ViewChild('chart') chartRef: ElementRef<SVGElement> = {} as ElementRef;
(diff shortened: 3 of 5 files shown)


src/app/features/charts/components/score-barchart/score-barchart.component.scss
```diff
@@ -4,6 +4,10 @@ $tooltip-color: #63676C;
     position: relative;
     height: 350px;
 
+    svg {
+        width: 100%; 
+    }
```
hallieswan, line 9: Ensure chart takes up the full available width when initially rendered

src/app/features/charts/components/score-barchart/score-barchart.component.html
```diff
@@ -1,7 +1,7 @@
-<div id="score-barchart">
+<div id="score-barchart" #div (window:resize)="onResize(div.getBoundingClientRect().width)">
```
hallieswan: Get the width of the div when the window is resized, so the chart can also be resized

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -324,4 +357,56 @@ export class ScoreBarChartComponent implements OnChanges, AfterViewInit, OnDestr
     if (this.initialized) 
       this.chart.remove();
   }
+
+  @HostListener('window:resize', ['$event'])
+  onResize(divSize: number) {
```
hallieswan: Based on existing method in [base-chart](https://github.com/Sage-Bionetworks/Agora/blob/66d57a3f4f56402b45e3b72c0d5a949c499c0c93/src/app/features/charts/components/base-chart/base-chart.component.ts#L116)

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -324,4 +357,56 @@ export class ScoreBarChartComponent implements OnChanges, AfterViewInit, OnDestr
...
+    if (this.shouldResize && this.initialized) {
+      const self = this;
+      clearTimeout(this.resizeTimer);
+      this.resizeTimer = setTimeout(() => {
+        self.resizeChart(divSize);
+      }, 100); 
+    }
+  };
+
+  resizeChart = (divSize: number): void => {
```
hallieswan, line 375: Resizes the chart width with a transition when the window is resized

src/app/features/genes/components/gene-soe-charts/gene-soe-charts.component.scss
```diff
@@ -4,15 +4,12 @@
 @import '../../../../../styles/mixins';
 
 .gene-score-charts {
-  display: flex;
-  flex-direction: row;
-  flex-wrap: wrap;
-  justify-content: center;
-  gap: 20px;
+  display: grid;
```
hallieswan, line 7: Use a grid to ensure that all plots have the same width, even when wrapped onto multiple rows

src/app/features/charts/components/score-barchart/score-barchart.component.html
sagely1: As a suggestion, perhaps we can rename #div to #scoreBarChart.
We can also move the width logic out of the html and have the onResize method calculate the div width. 
 Then we have just presentation logic in the html.
Then we can do something like this in the html:
`<div id="score-barchart" #scoreBarChart>`

And in the component:
```
@HostListener('window:resize', ['$event.target'])
onResize() {
   const divWidth = this.scoreBarChart.getBoundingClientRect().width;
   ...
}
```

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -28,6 +28,7 @@ export class ScoreBarChartComponent implements OnChanges, AfterViewInit, OnDestr
     this._score = score;
   }
 
+  @Input() shouldResize = true;
```
sagely1, line 31: nice, I like this option

src/app/features/charts/components/score-barchart/score-barchart.component.ts
```diff
@@ -121,22 +132,29 @@ export class ScoreBarChartComponent implements OnChanges, AfterViewInit, OnDestr
...
+        .width || this.MIN_CHART_WIDTH
+    );
+  }
+
   createChart() {
     this.initData();
     if (this.chartData) {
-      const width = 350;
+      const width = this.getChartBoundingWidth();
       const height = 350;
```
sagely1: Perhaps we also pull height out of this method, so we have all of our variables at the top?

src/app/features/charts/components/score-barchart/score-barchart.component.ts
sagely1: As a suggestion, perhaps we can just move this code up near the top?  I think this might be a general convention.

sagely1 (commented):
Left you some feedback.  Let me know if you have questions!

src/app/features/charts/components/score-barchart/score-barchart.component.html
hallieswan: Sure thing, done in the next commit

src/app/features/charts/components/score-barchart/score-barchart.component.ts
hallieswan: Sure! Done in the next commit

sagely1 (approved):
Looks good

PR #1205, opened 2023-06-26
AG-1115: prevent CI values from overlapping in GCT circle overlay by re-positioning labels

hallieswan:
gene tissue | develop | feature
:---:|:---:|:---:
ABCA8 PHG | develop_ABCA8_PHG | feature_ABCA8_PHG_moveLabels
APP CBE | develop_APP_CBE | feature_APP_CBE_moveLabels
APP DLPFC | develop_APP_DLPFC | feature_APP_DLPFC_moveLabels

### DIFF
--- a/src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.scss
+++ b/src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.scss
@@ -134,7 +134,14 @@
           font-size: var(--font-size-sm);
           top: 100%;
           left: 50%;
-          transform: translate(-50%, 14px);
+          
+          &.gct-details-panel-chart-interval-left {
+            transform: translate(-95%, 14px);
+          }
+
+          &.gct-details-panel-chart-interval-right {
+            transform: translate(-5%, 14px);
+          }
         }
 
         &:first-child {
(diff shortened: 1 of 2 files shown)


src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.scss
```diff
@@ -134,7 +134,14 @@
           font-size: var(--font-size-sm);
           top: 100%;
           left: 50%;
-          transform: translate(-50%, 14px);
+          
+          &.gct-details-panel-chart-interval-left {
+            transform: translate(-95%, 14px);
+          }
```
hallieswan, line 140: Align the last digit of the left interval label with the tick, rather than centering the interval label

src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.scss
```diff
@@ -134,7 +134,14 @@
...
+
+          &.gct-details-panel-chart-interval-right {
+            transform: translate(-5%, 14px);
+          }
```
hallieswan, line 144: Align the first digit of the right interval label with the tick, rather than centering the interval label

sagely1 (approved):
I think this works great as a very straightforward, simple and elegant way to address overlaps.  A more complex approach could use something like calc() to determine whether there would be overlap and dynamically adjust the percentages depending on the spacing of the two values.  Another approach we discussed is changing the range to scale the chart.  We will leave it up to @JessterB to weigh in.  Thanks!

JessterB (approved):
This is the winner!

PR #1206, opened 2023-06-27
AG-1115: prevent CI values from overlapping in GCT circle overlay by adjusting displayed limits

hallieswan:
gene tissue | develop | feature
:---:|:---:|:---:
ABCA8 PHG | develop_ABCA8_PHG | feature_ABCA8_PHG_adjustLimits
APP CBE | develop_APP_CBE | feature_APP_CBE_adjustLimits
APP DLPFC | develop_APP_DLPFC | feature_APP_DLPFC_adjustLimits

### DIFF
--- a/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.helpers.ts
+++ b/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.helpers.ts
@@ -114,7 +114,7 @@ export const getDetailsPanelData = function (
     }
   });
 
-  max = Math.ceil(max);
+  max = max * 1.1;
 
   const data = {
     gene: gene,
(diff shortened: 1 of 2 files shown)


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.helpers.ts
```diff
@@ -114,7 +114,7 @@ export const getDetailsPanelData = function (
     }
   });
 
-  max = Math.ceil(max);
```
hallieswan, line 117: Limits are still calculated using the max value for all tissues for a given gene and are still centered around 0, but scaling the max value slightly rather than rounding to the nearest integer ensures that tissues with small values are more clearly displayed

sagely1 (approved):
I personally like this approach better but if we need to keep the range consistent, we will have to go with the other approach.  We will let @JessterB weight in.

hallieswan:
We decided to reposition the labels (see #1205) instead of adjusting the limits, so closing this PR

PR #1211, opened 2023-07-05
AG-1170: fix url parameter for p-value filter

hallieswan:
- Use "significance" instead of "significanceThreshold" for the p-value filter url parameter


JessterB (approved):
LGTM!

PR #1213, opened 2023-07-05
AG-1166: title of go terms section doesn't display biodomain when there are no go terms

hallieswan:
https://github.com/Sage-Bionetworks/Agora/assets/26949006/ec2b3cbc-849a-4d5d-9a0b-cf78e0501273


JessterB (approved):
nice fix!

PR #1216, opened 2023-07-07
AG-1113: use preventDefault to prevent score labels from moving when double clicking on histogram in GCT overlay

hallieswan:
Downside of this approach -- the context menu no longer appears when right clicking on the label

https://github.com/Sage-Bionetworks/Agora/assets/26949006/199169bb-2960-4e51-8fe9-78c5b1027f6d


sagely1 (commented):
You can close this one.  I like the other approach!

PR #1217, opened 2023-07-07
AG-1113: use user-select to prevent score labels from moving when double clicking on histogram in GCT overlay

hallieswan:
Pro: context menu still appears when right clicking the label
Requires: updating caniuse-lite so stylelint plugin/no-unsupported-browser-features recognizes that "user-select-none" is supported by UC Browser for Android

https://github.com/Sage-Bionetworks/Agora/assets/26949006/48d287b3-6f82-4f38-9dca-317d633ebd1c

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5779,9 +5779,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001380",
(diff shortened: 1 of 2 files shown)


sagely1 (approved):
Looks good!

PR #1229, opened 2023-08-22
AG-1173 added TEP logic to gene hero

sagely1:


src/app/features/genes/components/gene-hero/gene-hero.component.spec.ts
```diff
@@ -28,12 +29,33 @@ describe('Component: Gene Hero', () => {
...
+
+  it('should only add TEP text if nominations is null and either is_tep or is_adi is true', () => {
+    component.gene = geneMock3;
+    fixture.detectChanges();
+    
+    const el = element.querySelector('.gene-hero-nominated') as HTMLElement;
+    
+    const expected = 'Selected for Target Enabling Resource Development';
+    expect(el.textContent).toBe(expected);
+  });
```
hallieswan: Nice tests! You could also consider adding tests to show that 1) only "Nominated Target" is shown when is_tep and is_adi are both false, and 2) nothing is displayed when nominations is null and both is_tep and is_adi are false. Up to you though!

src/app/testing/gene-mocks.ts
```diff
@@ -3144,6 +3146,137 @@ export const geneMock2: Gene = {
     },
   ],
   nominations: 4,
+  is_adi: false,
+  is_tep: true
+};
+
+export const geneMock3: Gene = {
```
hallieswan, line 3153: From [the build log](https://app.travis-ci.com/github/Sage-Bionetworks/Agora/builds/265394612), it looks like this mock might be missing `nominatedtarget` and `nominations` fields:

```log
ERROR in src/app/testing/gene-mocks.ts:3153:14 - error TS2739: Type '{ _id: string; ensembl_gene_id: string; name: string; 
summary: string; hgnc_symbol: string; alias: string[]; isIGAP: false; haseqtl: true; isAnyRNAChangedInADBrain: true; 
rna_brain_change_studied: true; ... 5 more ...; is_tep: true; }' is missing the following properties from type 'Gene':
nominatedtarget, nominations

3153 export const geneMock3: Gene = {
                  ~~~~~~~~~
```

hallieswan (commented):
Looks good! Just looks like the new mock may need to be tweaked

src/app/testing/gene-mocks.ts
sagely1: Looks like I didn't commit all my changes.  Good catch!

src/app/features/genes/components/gene-hero/gene-hero.component.spec.ts
sagely1: My latest commit should have these missing tests.  thanks!

hallieswan (approved):
Looks good! Could also add a test to check when `nominations` is null and both `is_adi` and `is_tep` are false, but up to you!

sagely1:
> Looks good! Could also add a test to check when `nominations` is null and both `is_adi` and `is_tep` are false, but up to you!

Added border condition unit test as per @hallieswan

PR #1230, opened 2023-08-24
AG-1152 new TREAT-AD resource section

sagely1:


src/app/features/genes/components/gene-resources/gene-resources.component.ts
```diff
@@ -1,38 +1,23 @@
...
-  get gene(): Gene | undefined {
-    return this._gene;
-  }
-  @Input() set gene(gene: Gene | undefined) {
-    this._gene = gene;
-    this.init();
-  }
-  additionalResources: AdditionalResource[] = [] as AdditionalResource[];
+export class GeneResourcesComponent implements OnInit {
+  @Input() gene: Gene | undefined;
```
sagely1, line 11: standardizing inputs

src/app/models/genes.ts
```diff
@@ -69,6 +69,7 @@ export interface Gene {
   nominations: number | null;
   is_adi: boolean;
   is_tep: boolean;
+  resource_url: string;
```
sagely1: new property

src/app/features/genes/components/gene-resources/gene-resources.component.ts
```diff
@@ -1,38 +1,23 @@
...
+  
+  additionalResources: AdditionalResource[] = [];
 
-  reset() {
-    this.additionalResources = [];
+  ngOnInit(): void {
+    this.init();
```
sagely1, line 16: moved code to ngOnInit() as using the setter to update the static content on changes didn't seem to be needed.

src/app/features/genes/components/gene-resources/gene-resources.component.html
```diff
@@ -20,12 +20,44 @@ <h2>
...
+          </div>
+          <div class="col-md-6 col-sm-12 col-12 header-description">View the openly available TREAD-AD resources for experimental validation of {{ gene.hgnc_symbol || gene.ensembl_gene_id }} in the AD Knowledge Portal.</div>
+        </div>
+        <div class="row tab-row last-row">
+          <div class="col-md-3 col-sm-8 col-8 header-title">Target Portfolio</div>
+          <div class="col-md-3 col-sm-4 col-4 text-center header-link">
+            <a class="link no-bold" href="https://treatad.org/data-tools/target-dashboard" target="_blank">View Resource Status</a>
+          </div>
+          <div class="col-md-6 col-sm-12 col-12 header-description">View the status of TEP resource development on the TREAT-AD Target Portfolio and Progress Dashboard.</div>
+        </div>
```
hallieswan: The ticket states that card 2 "is displayed only for genes with `is_tep == TRUE`" -- should an `*ngIf` be added here?

src/app/features/genes/components/gene-resources/gene-resources.component.html
```diff
@@ -20,12 +20,44 @@ <h2>
...
+        <div *ngIf="gene.is_adi" class="row tab-row last-row">
+          <div class="col-md-3 col-sm-8 col-8 header-title">AD Informer Set</div>
+          <div class="col-md-3 col-sm-4 col-4 text-center header-link">
+            <a class="link no-bold" href="https://treatad.org/data-tools/ad-informer-set" target="_blank">View Resource Status</a>
+          </div>
```
hallieswan: In the ticket, it looks like card 3 should have the link text "View AD Informer Set" -- should this text be updated?

src/app/features/genes/components/gene-resources/gene-resources.component.spec.ts
```diff
@@ -8,28 +8,78 @@ import { RouterTestingModule } from '@angular/router/testing';
...
+    component.gene.is_tep = false;
+
+    fixture.detectChanges();
+
+    const el = element.querySelector('#target-enabling-resources-header');
+    
+    expect(el).toBe(null);
+  });
+
+  it('should display TREAT-AD resource section if is_tep or is_adi is true', () => {
```
hallieswan: Nice tests! You could also consider adding tests to demonstrate that card 2 and card 3 are conditionally shown based on `is_tep` and `is_adi` values respectively.

src/app/testing/gene-mocks.ts
```diff
@@ -3278,7 +3280,8 @@ export const geneMock3: Gene = {
   ],
   nominations: null,
   is_adi: false,
-  is_tep: true
+  is_tep: true,
+  resource_url: ''
```
hallieswan: Is an empty string an acceptable value for `resource_url` when `is_tep` is true? Just curious based on this statement in the ticket: "When the above condition is satisfied, gene_info.resource_url should always be populated." where the condition is that either `is_tep` or `is_adi` is true.

src/app/models/AdditionalResource.ts
```diff
@@ -0,0 +1,6 @@
+export interface AdditionalResource {
```
hallieswan, line 1: Could consider using this interface to structure the data for the new TEP resource cards?

hallieswan (changes requested):
A couple questions based on the ticket, but looking good!

src/app/models/AdditionalResource.ts
sagely1: It is implemented in gene-resources.component.ts.  Let me know if I misunderstood.

src/app/testing/gene-mocks.ts
sagely1: You're right.  I made a change to the original interface since it could be null.

src/app/models/genes.ts
sagely1: updated the type of this to a nullable string

src/app/features/genes/components/gene-resources/gene-resources.component.ts
```diff
@@ -1,38 +1,25 @@
...
+  additionalResources: AdditionalResource[] = [];
+
+  ngOnInit(): void {
     this.init();
-  }
-  additionalResources: AdditionalResource[] = [] as AdditionalResource[];
 
-  reset() {
-    this.additionalResources = [];
+    console.log(this.gene?.resource_url);
```
hallieswan: Is this console.log statement still needed?

src/app/testing/gene-mocks.ts
```diff
@@ -3278,7 +3280,8 @@ export const geneMock3: Gene = {
   ],
   nominations: null,
   is_adi: false,
-  is_tep: true
+  is_tep: true,
+  resource_url: null
```
hallieswan, line 3284: Thanks for allowing the null type for `resource_url`! I'm still curious about the data in this mock though -- since `is_tep` is true, should `resource_url` contain a URL, given the data validation rules described in the ticket? Alternately, `is_tep` could be set to false and `resource_url` maintained as null. What do you think?

For reference, here is the data validation rule described in the ticket: 

> If the data has TRUE boolean(s) but no resource_url, that is a data issue. You shouldn’t see that, as I have validation rules in place to detect it.

Where the booleans referred to are: 

>   - gene_info.is_tep == TRUE
>   - gene_info.is_adi == TRUE

hallieswan (approved):
A few clarifying questions, but looks good!

src/app/features/genes/components/gene-resources/gene-resources.component.ts
sagely1: good catch! let me remove that.

src/app/testing/gene-mocks.ts
sagely1: I think since this is a scenario that shouldn't occur upstream, I was mostly concerned that it the value was null, the page would at least continue to load.  I focused on making sure the cards didn't show inadvertently.  Ideally, I would want to log and monitor for data issues in the future as well as incorporating some data quality checks, but I consider these items as a future enhancement.  Let me know if that makes sense!

src/app/models/AdditionalResource.ts
hallieswan: `AdditionalResource` is used to structure the data for the additional resources displayed below the TREAT-AD resource section. I was wondering if the same interface could be used to structure the new TEP resource cards, so it would be easier to see the card data separate from the html tags. For example, set up the card data in the component: 

```typescript
// gene-resources.component.ts
const card1 = {
  title: 'Target Enabling Resources',
  description:
  `View the openly available TREAT-AD resources for experimental validation of ${this.gene?.gene.hgnc_symbol || this.gene?.ensembl_gene_id} in the AD Knowledge Portal.`,
  linkText: 'View Available Resources',
  link: `${this.gene?.resource_url}`,
 }
const card2 = {...}
const card3 = {...}
```
Then `card1`, `card2`, and `card3` could be used in the template, similar to how `additionalResources` is used. 

However, all good to keep things as they are! Up to you.

src/app/testing/gene-mocks.ts
hallieswan: Ah okay, sounds good!

hallieswan (approved):

src/app/models/AdditionalResource.ts
sagely1: Great suggestion considering the similarities.  I think for now, I will leave these as is since the TEP resources are driven off the properties of the input gene object.

PR #1236, opened 2023-09-26
AG-1214 update the url param behavior for model dropdown

sagely1:
Bug revealed a change required to how we handle urls when navigating between subtabs in the gene details page.  Since we don't have specifics on a url sharing strategy as of yet, we will add a url parameter for the model in the url.  If the page is refreshed, the page will use the url parameter (if it exists) to preselect a dropdown option on the RNA page.  Since navigation can occur between subtabs, we will keep the url parameter and not remove it (as it was before).


src/app/core/services/helper.service.ts
```diff
@@ -197,7 +197,7 @@ export class HelperService {
     const param = new HttpParams()
       .set(paramName, paramValue);
     console.log(param.toString());
-    return `${url}?${param.toString()}`;
+    return `${url.split('?')[0]}?${param.toString()}`;
```
sagely1, line 199: although we don't support multiple url parameters here yet, updating the url parameter is supported with this logic

src/app/core/services/helper.service.ts
```diff
@@ -192,6 +193,13 @@ export class HelperService {
     return urlParams.get(name);
   }
 
+  addUrlParam(url: string, paramName: string, paramValue: string | number | boolean) {
```
sagely1: this method adds the ability to generate a url with a single parameter (so we can support adding the 'model' url param)

src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
```diff
@@ -194,9 +193,10 @@ export class GeneEvidenceRnaComponent implements AfterViewInit {
   }
 
   onStatisticalModelChange(event: any) {
-    if (!this._gene?.rna_differential_expression) {
+    if (!event)
```
sagely1, line 196: original source of error in PR.  Updated method to be more defensive about inputs

src/app/features/genes/components/gene-model-selector/gene-model-selector.component.ts
```diff
@@ -38,7 +38,11 @@ export class GeneModelSelectorComponent {
   ngOnInit() {
     this.route.queryParams.subscribe((params) => {
       const modelFromURL = params['model'];
-      const index = this._options.findIndex(o => o.value === modelFromURL);
+      let index = this._options.findIndex(o => o.value === modelFromURL);
+      if (index === -1) {
+        // default to first option if page is loaded without a model parameter
+        index = 0;
```
sagely1, line 44: although this shouldn't be the case when navigating, if the page should be opened without a parameter, we will default to the first option as per @JessterB

src/app/core/services/helper.service.ts
```diff
@@ -115,7 +115,7 @@ export class HelperService {
     }
   }
 
-  getOffset(el: any) {
+  getOffset(el: Element) {
```
sagely1: will eventually strongly type, but I will revert this now as there are dc.js charts using this.  When dc.js is completely removed eventually, this and other helper methods will be revisited.

src/app/core/services/helper.service.ts
hallieswan: Makes sense! What do you think about adding a comment to capture this constraint or updating the name, e.g. `addSingleUrlParam`?

src/app/core/services/helper.service.spec.ts
```diff
@@ -146,4 +146,18 @@ describe('Service: Helper', () => {
...
+    const result = helperService.addUrlParam(url, 'model', 'abc');
+    expect(result).toEqual(expected);
+  });
+
+  it('should replace existing url parameters if the same one exists', () => {
+    const url = '/genes/ENSG123?model=abc';
+    const expected = '/genes/ENSG123?model=fff';
+    const result = helperService.addUrlParam(url, 'model', 'fff');
+    expect(result).toEqual(expected);
+  });
```
hallieswan: Nice! Could you add another test that demonstrates the expected behavior when the URL contains an existing, different query parameter? For example, if the url is `/genes/ENSG123?existing=abc` and we call `helperService.addUrlParam(url, 'model', 'fff')`, what should the result be?

src/app/features/genes/components/gene-model-selector/gene-model-selector.component.ts
hallieswan: Will the url be updated to include the model query param in this case?

src/app/core/services/helper.service.ts
sagely1: good suggestion.  I will rename it.

src/app/features/genes/components/gene-model-selector/gene-model-selector.component.ts
sagely1: I decided not to since it's not an expected scenario and we will default to the first option in this case.

src/app/core/services/helper.service.spec.ts
sagely1: Good question.  It should replace the url parameter as this should be the only url parameter as of this point.  I'll add the unit test to add this condition.

hallieswan (approved):
Looks good to me!

PR #1238, opened 2023-09-26
AG-300 updated json keys

sagely1:


src/app/testing/gene-mocks.ts
```diff
@@ -90,101 +86,100 @@ export const geneMock1: Gene = {
       initial_nomination: 2018,
     },
   ],
-  medianexpression: [
+  median_expression: [
     {
       ensembl_gene_id: 'ENSG00000147065',
```
hallieswan: The ticket notes that `medianExpression.ensembl_gene_id` was removed from `median_expression` -- should these mocks be updated to remove`ensembl_gene_id` from `median_expression`?

src/app/models/genes.ts
```diff
@@ -29,17 +28,16 @@ export interface NominatedTarget {
 
 export interface MedianExpression {
   ensembl_gene_id: string;
```
hallieswan: Should `ensemble_gene_id` also be removed from this interface?

sagely1: Good catch!  I only removed it on the api side.

src/app/testing/gene-mocks.ts
sagely1: yes, intellisense will demand it. 👍

hallieswan (approved):
Looks good to me!

PR #1239, opened 2023-09-27
AG-1208 update PUB AD resource card link

sagely1:


src/app/features/genes/components/gene-resources/gene-resources.component.ts
```diff
@@ -62,7 +62,7 @@ export class GeneResourcesComponent implements OnInit {
         description:
           'View dementia-related publication information for this gene on PubAD.',
         linkText: 'Visit PubAD',
-        link: `https://adexplorer.medicine.iu.edu/pubad/external/${this.gene?.ensembl_gene_id}`,
+        link: 'https://adexplorer.medicine.iu.edu/pubad',
```
hallieswan: Would it be possible to create the link with a method, rather than updating the link after creating `this.additionalResources`? e.g. something like this -- 

```TS
function createPubAdLink(hgnc_symbol) {
  const baseUrl = 'https://adexplorer.medicine.iu.edu/pubad'
  return hgnc_symbol ? `${baseUrl}/external/${hgnc_symbol}` : baseUrl
}

this.additionalResources = [
....
        link: createPubAdLink(this.gene?.hgnc_symbol)
....
```

sagely1: Unfortunately, I don't think you can do that within an array assignment.

hallieswan (approved):
Looks good to me, just one small question!

src/app/features/genes/components/gene-resources/gene-resources.component.ts
sagely1: Good idea.  I can write the function and call it inside of the assignment: ${ createPubAdLink() }

hallieswan (approved):

PR #1245, opened 2023-10-16
AG-1205 pinning peptides bug

sagely1:
Pinning peptides when "PROTEIN - DIFFERENTIAL EXPRESSION" was selected resulted in variants being removed incorrectly due to a missing filter.


hallieswan (approved):
Looks good to me! Could also add a test case to show that the bug is resolved, but up to you!

sagely1:
> Looks good to me! Could also add a test case to show that the bug is resolved, but up to you!

Since I will be removing Cypress, I will add some e2e tests for this scenario.

PR #1247, opened 2023-10-18
AG-1237 add Target Enabling Resources filter to GCT

sagely1:


src/server/components/comparison.ts
```diff
@@ -125,12 +125,22 @@ function getComparisonGene(gene: Gene, teams: Team[], scores: Scores[], allBiodo
...
     multi_omics_score: geneScores ? geneScores.multi_omics_score : null,
-    biodomains: geneBiodomains
+    biodomains: geneBiodomains,
+    target_enabling_resources: getTargetEnablingResources(gene)
   };
 
   return data;
 }
 
+export function getTargetEnablingResources(gene: Gene) {
```
sagely1, line 135: incoming data is booleans but we need a string array for filtering, so this function will create the needed string array.

hallieswan (approved):
Looks good to me! Just curious, I noticed that the ticket noted some complexity around handling filtering for multiple rows in the proteomics view -- is that something that was already handled by the GCTFilter?

sagely1:
> Looks good to me! Just curious, I noticed that the ticket noted some complexity around handling filtering for multiple rows in the proteomics view -- is that something that was already handled by the GCTFilter?

If you are referring to the boolean AND/OR logic, that is the default behavior.  It was just being explicitly stated in the ticket.

PR #1250, opened 2023-10-25
AG-1215 removed cypress and added playwright

sagely1:
removed cypress dependencies and added playwright e2e testing


package.json
```diff
@@ -39,8 +38,7 @@
     "mongo:start": "sh ./scripts/start-mongo.sh '/var/lib/mongodb/'",
     "mongo:start:windows": "sh ./scripts/start-mongo.sh 'C:\\data\\db'",
     "data:local": "npm run clean:data && sh ./scripts/get-data-local.sh",
-    "cypress": "cypress",
-    "cypress:run": "cypress run --config-file ./config/cypress.config.js"
+    "e2e": "playwright test"
```
hallieswan: Should the `e2e` script start the server, like the previous `test:e2e` script? Alternately, the [`webServer`](https://playwright.dev/docs/test-webserver) in the `playwright.config` could be updated to launch the server, if it isn't already running.

playwright.config.ts
```diff
@@ -0,0 +1,77 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    // baseURL: 'http://127.0.0.1:3000',
```
hallieswan: Can this be updated to specify a `baseURL`? Playwright [recommends](https://playwright.dev/docs/test-webserver#adding-a-baseurl) adding a `baseURL`, so that tests can use relative URLs. 

```suggestion
    /* Base URL to use in actions like `await page.goto('/')`. */
    baseURL: 'http://127.0.0.1:8888',
```

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
hallieswan, line 1: Should this file be included, if the e2e tests are run in the Travis workflow? Just wondering if we're intending to run the e2e tests in Travis and in GitHub actions

tests/homepage.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  test('has news', async ({ page }) => {
+    await page.goto('http://localhost:8080');
+
+    // Hamburger menu should be hidden for the given viewport above
+    expect(page.locator('button.header-nav-toggle')).toBeHidden();
+    
+    // look for news link and click it
+    await page.locator('a span', { hasText: 'News' }).click();
```
hallieswan: Using the `locator` directly works great, but could also consider using the [Playwright recommended built-in locator](https://playwright.dev/docs/locators#quick-guide): 

```suggestion
    // look for news link and click it
    await page.getByRole('link', { hasText: 'News' }).click();
```

tests/homepage.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('specific viewport block', () => {
+  test.use({ viewport: { width: 1600, height: 1200 } });
+
+  test('has title', async ({ page }) => {
+    await page.goto('http://localhost:8080');
```
hallieswan: If a `baseURL` is added to the playwright.config, could update these calls to use relative urls: 
```suggestion
    await page.goto('/');
```

hallieswan (commented):
Looking great! Just a few questions for you

playwright.config.ts
sagely1: Great suggestion! I'll add a base url.

.github/workflows/playwright.yml
sagely1: Since the plan is to move to GitHub actions, my plan is to just keep this as a placeholder for now.  It won't execute since I made sure to not configure the branch names until we are ready to go.  Once it's working I'll remove Travis.

tests/homepage.spec.ts
sagely1: Implemented :)

package.json
sagely1: Great idea! One observation was that I found instances where changing the playwright config would result in different results when rerunning tests and this lead me to question whether I need to restart the "app" and "server" instances for the new config to take.  For now, I think I'll just mimic the previous cypress behavior but I think I will revisit this once I dial in the playwright config.  Let me know if you had a similar experience.

tests/homepage.spec.ts
sagely1: I guess this is a mind shift since I wanted to be explicit about the elements used.  I guess the idea is to approach testing with accessibility in mind.  From the docs: "We recommend using text locators to find non interactive elements like div, span, p, etc. For interactive elements like button, a, input, etc. use [role locators](https://playwright.dev/docs/locators#locate-by-role)".

I ended up using the getByRole and name attributes to match the documentation.  I wasn't sure why Playwright  uses 'name' instead of something like 'text' in the docs since html elements can have a 'name' attribute and that could be more confusing.

package.json
hallieswan: Oh interesting! I haven't run into that issue before, but makes sense to keep it as is for now

hallieswan (approved):
Looks great! Super exciting to see Playwright used in Agora

PR #1255, opened 2023-11-01
AG-1278 fix nominations ordering

sagely1:


JessterB (changes requested):
This looks good, but I did notice a small discrepancy in geneMock1 (MSN). The bject has 5 items in target_nominations, but "total_nominations: 4" - those should match.

I don't think that actually impacts this test since 5|4 > 3 > 0, but we should align those values in case we do add a test where that matters.

JessterB (approved):

PR #1259, opened 2023-11-06
AG-1289 disabled e2e tests until playwright config fixed

sagely1:


JessterB (approved):
AG-1294 tracks reenabling these

PR #1267, opened 2023-12-07
AG-959 gene hero surface ensembl version

sagely1:


.github/workflows/node.js.yml
```diff
@@ -5,18 +5,19 @@ name: Node.js CI
 
 on:
```
sagely1, line 6: You can ignore the review on this file.  I am waiting on some changes from IT before this starts working.

src/app/models/genes.ts
```diff
@@ -98,6 +99,8 @@ export interface Gene {
   input_data_display_value?: string;
 
   bio_domains?: BioDomains;
+
+  ensembl_info: EnsemblInfo[];
```
hallieswan, line 104: Is `ensembl_info` an array of `EnsemblInfo` in the database? Just curious, since the ticket seems to specify that `ensembl_info` is just an object of `EnsemblInfo`

src/app/features/genes/components/gene-hero/gene-hero.component.html
```diff
@@ -14,17 +14,31 @@ <h2 class="gene-hero-name">
...
+      <ng-container *ngIf="gene.bio_domains || getAlias() || getEnsemblUrl() !== ''">
         <hr />
         <div *ngIf="gene.bio_domains" class="gene-hero-biodomains">
           <h4 class="gene-hero-biodomains-heading">Biological Domains</h4>
           <p>
             {{ getBiodomains() }}
           </p>
         </div>
-        <div *ngIf="getAlias()" class="gene-hero-aliases">
+        <div *ngIf="getAlias() || getEnsemblUrl() !== ''" class="gene-hero-aliases">
```
hallieswan: Per the ticket, "The AKA header and the first line underneath (which displays the current gene details page’s ENSG)  it should always be displayed." Should the conditional be removed from this div?

src/app/features/genes/components/gene-hero/gene-hero.component.html
```diff
@@ -14,17 +14,31 @@ <h2 class="gene-hero-name">
...
+          <ng-container *ngIf="getEnsemblUrl() === ''">
+            <span *ngIf="getEnsemblUrl() === ''">{{ gene.ensembl_gene_id }}</span>
+          </ng-container>
+          <p *ngIf="gene.ensembl_info[0].ensembl_possible_replacements.length > 0">
+            <a [href]="getPossibleReplacementsURL()">Possible replacement value<span *ngIf="gene.ensembl_info[0].ensembl_possible_replacements.length > 1">s</span></a>:
+            {{ gene.ensembl_info[0].ensembl_possible_replacements.join(', ')}}
+          </p>
           <p>
             {{ getAlias() }}
           </p>
```
hallieswan: Should a conditional be added to the `getAlias()` paragraph, since the parent div is no longer conditionally shown based on whether `getAlias()` exists?

src/app/features/genes/components/gene-hero/gene-hero.component.spec.ts
```diff
@@ -110,4 +110,14 @@ describe('Component: Gene Hero', () => {
     const expected = 'Immune Response, Lipid Metabolism, Structural Stabilization, Synapse, Vasculature';
     expect(component.getBiodomains()).toBe(expected);
   });
+
+  it('should return the ensembl permalink', () => {
+    component.gene = geneMock1;
+    expect(component.getEnsemblUrl()).toBe('https://may2015.archive.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000264794');
+  });
```
hallieswan, line 117: What do you think about adding a test case for a gene without an `ensembl_permalink`?

hallieswan (commented):
Looks good! A couple of questions. 

Also, what would you think about adding a Playwright test for the various conditions? The gene data could be [mocked](https://playwright.dev/docs/mock#mock-api-requests)!

src/app/models/genes.ts
sagely1: It is an array in MongoDB but that is a bug.  This will be fixed in a future data release and will require some code changes to match.

src/app/features/genes/components/gene-hero/gene-hero.component.spec.ts
sagely1: good idea, i'll add that.

src/app/features/genes/components/gene-hero/gene-hero.component.html
sagely1: you're right.  it is redundant.

src/app/features/genes/components/gene-hero/gene-hero.component.html
sagely1: correct.  I fixed the parent div and added the condition

src/app/models/genes.ts
hallieswan: Got it! Would it be useful to add a FIXME comment here with the associated ticket number, so it's easier to track the fix in the future?

hallieswan (approved):
Looks good to me!

src/app/models/genes.ts
sagely1: Added!

PR #1274, opened 2024-01-08
AG-1281 fixing obscured tooltip

sagely1:
Fixing prime dropdown list items obscured when user opens prime dropdown and then hovers over a tooltip-enabled element.

### DIFF
--- a/src/styles/components/_dropdown.scss
+++ b/src/styles/components/_dropdown.scss
@@ -32,6 +32,7 @@ div.p-dropdown {
 }
 
 .p-multiselect {
+  z-index: 1000 !important; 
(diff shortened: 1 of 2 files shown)


src/styles/components/_dropdown.scss
```diff
@@ -32,6 +32,7 @@ div.p-dropdown {
 }
 
 .p-multiselect {
+  z-index: 1000 !important;
```
sagely1, line 35: straightforward fix as z-index for the multiselect just needs to be higher than the tooltip

hallieswan (approved):

PR #1276, opened 2024-01-16
AG-1222 add SRM results to gene details

sagely1:


src/app/features/genes/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
```diff
@@ -74,10 +78,77 @@ export class GeneEvidenceProteomicsComponent {
...
+
+          SRMData.push({
+            key: data.tissue,
+            value: [data.min, data.median, data.max],
+            circle: {
+              value: item.log2_fc,
+              tooltip:
+                (item.hgnc_symbol || item.ensembl_gene_id) +
+                ' is ' +
+                (item.cor_pval <= 0.05 ? ' ' : 'not ') +
```
hallieswan: If the p-value is less than 0.05, should an empty string be used so there isn't an extra space in the sentence?

src/app/features/genes/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
```diff
@@ -74,10 +78,77 @@ export class GeneEvidenceProteomicsComponent {
       this.selectedUniProtId = this.uniProtIds[0];
     }
 
+    this.initSRM();
+
     this.initLFQ();
     this.initTMT();
   }
 
+  initSRM() {
```
hallieswan, line 113: `initSRM` looks similar to `initLFQ` and `initTMT` -- would it be possible to extract the shared logic into a helper method?

hallieswan (approved):
Looks good to me! Just a couple small questions

src/app/features/genes/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
sagely1: Nice find.  I will remove it and refactor this as it is used in multiiple places.

src/app/features/genes/components/gene-evidence-proteomics/gene-evidence-proteomics.component.ts
sagely1: Yes, I'll do the refactor as all three seem to share the same logic.

PR #1278, opened 2024-01-22
AG-1221 add SRM option to GCT subcategory dropdown options

sagely1:


src/server/api.ts
```diff
@@ -38,23 +38,9 @@ const database = { url: '' };
         doc && ('doc => ' + util.inspect(doc)), '\n');
 }); */
 
-console.log('HELLLLLLOOOOOOOO');
```
sagely1, line 41: unrelated to the PR but is extraneous

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.variables.ts
```diff
@@ -32,13 +32,17 @@ export const subCategories: { [key: string]: GCTSelectOption[] } = {
   ],
   'Protein - Differential Expression': [
     {
-      label: 'Label-free Quantification (LFQ)',
-      value: 'LFQ',
+      label: 'Targeted Selected Reaction Monitoring (SRM)',
```
sagely1, line 35: adding option here

tests/gene-comparison-tool.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    test.slow();
+    // set category for Protein - Differential Expression
+    await page.goto('/genes/comparison?category=Protein+-+Differential+Expression');
+
+    // wait for page to load (i.e. spinner to disappear)
+    await expect(page.locator('div:nth-child(4) > div > .spinner'))
+      .not.toBeVisible({ timeout: 150000});
+  
+    // expect sub-category dropdown to be SRM
+    const dropdown = await page.locator('#subCategory');
```
sagely1: SRM should be the default option

src/server/components/comparison.ts
```diff
@@ -208,11 +209,16 @@ export async function getProteinComparisonGenes(method: string) {
...
     items = await ProteomicsLFQCollection.find()
       .lean()
       .sort({ hgnc_symbol: 1, tissue: 1 })
       .exec();
+  } else if ('SRM' === method) {
+    items = await ProteomicsSRMCollection.find()
+      .lean()
+      .sort({ hgnc_symbol: 1, tissue: 1 })
+      .exec();
   }
```
hallieswan: Would it be useful to include a final else block to handle unexpected values for method, e.g. log a message or throw an error?

tests/gene-comparison-tool.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    // wait for page to load (i.e. spinner to disappear)
+    await expect(page.locator('div:nth-child(4) > div > .spinner'))
+      .not.toBeVisible({ timeout: 250000});
+
+    // Expect a title "to contain" a substring.
+    await expect(page).toHaveTitle('Gene Comparison | Visual comparison tool for AD genes');
+  });
+
+  test('sub-category is SRM by default', async ({ page }) => {
+    test.slow();
```
hallieswan: `test.slow()` in `test.describe` should apply to all tests in the describe block, so this line may not be needed

tests/gene-comparison-tool.spec.ts
hallieswan: You may not need this await -- [page.locator](https://playwright.dev/docs/api/class-page#page-locator) isn't asynchronous

hallieswan (commented):
The changes to the menu and default order look good!

src/server/components/comparison.ts
sagely1: Excellent point. I've noticed that a lot of the existing codebase is currently just returning null/empty for these corner scenarios.  So until there is a better solution for capturing errors and alerting/logging, I think I will add an else + TODO for the time being.

tests/gene-comparison-tool.spec.ts
sagely1: removed!

tests/gene-comparison-tool.spec.ts
sagely1: You're right.  Removed!

hallieswan (commented):
The update to the header label in the circle overlay looks good! Do the row names in the GCT table also need to be updated [here](https://github.com/Sage-Bionetworks/Agora/blob/develop/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html#L315), so that the `uniprotid` is conditionally hidden when the SRM option is selected?

sagely1:
> The update to the header label in the circle overlay looks good! Do the row names in the GCT table also need to be updated [here](https://github.com/Sage-Bionetworks/Agora/blob/develop/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html#L315), so that the `uniprotid` is conditionally hidden when the SRM option is selected?

Good question!  I'll ask @JessterB

JessterB:
> > The update to the header label in the circle overlay looks good! Do the row names in the GCT table also need to be updated [here](https://github.com/Sage-Bionetworks/Agora/blob/develop/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html#L315), so that the `uniprotid` is conditionally hidden when the SRM option is selected?
> 
> Good question! I'll ask @JessterB

@sagely1 Yes, we do want to conditionally hide the `uniprotid` for SRM in the row names as well as in the overlay.

hallieswan (approved):
Looks good! Could consider adding a test to confirm that row names are displayed as expected for SRM vs other categories

PR #1282, opened 2024-02-12
AG-1346 GCT bug fix for zero p-values showing wrong circle size

sagely1:


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -912,15 +912,20 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
     const tissue = gene.tissues.find((t) => t.name === tissueName);
     let size = 0;
     let color = '#F0F0F0';
+    const CIRCLE_MAX_SIZE = 50;
 
     if (tissue?.adj_p_val) {
       size = this.getCircleSize(tissue.adj_p_val);
     }
 
+    if (size === 0) {
```
sagely1: if circle size is zero, it should have the max size of circle as the p-value is inversely proportional to the size of the circle.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -912,15 +912,20 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
+      size = CIRCLE_MAX_SIZE;
+    }
```
hallieswan: What do you think about handling this case in `getCircleSize` instead, so all of the sizing is determined in the same function?

sagely1: That's a good idea.  I refactored this method so there would be more clarity as there is too much logic spread out here for size and style.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -898,36 +901,35 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
+      return 0;
+
+    // if significance cutoff radio button selected and 
+    // p-Value > significance threshhold, don't show
     if (this.significanceThresholdActive && pval > this.significanceThreshold) {
       return 0;
     }
 
     const pValue = 1 - (this.nRoot(pval, 3) || 0);
-    const size = Math.round(100 * pValue * 0.44);
```
sagely1, line 907: 100 * 0.44 = 44 (What was previously the max size)
Since the legend is 50px for the largest circle, let's just multiply the pValue by the MAX_SIZE for simplicity.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -898,36 +901,35 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
-    return size < 6 ? 6 : size;
+    const size = Math.round(pValue * MAX_SIZE);
+
+    // ensure the smallest circles have a min size to be easily hoverable/clickable
+    return size < MIN_SIZE ? MIN_SIZE : size;
   }
 
   getCircleStyle(tissueName: string, gene: GCTGene) {
     const tissue = gene.tissues.find((t) => t.name === tissueName);
-    let size = 0;
```
sagely1, line 913: Methods are ideally small and clear so I refactored this for clarity

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -898,36 +901,35 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
-    if (tissue?.logfc) {
-      color = this.getCircleColor(tissue.logfc);
-    }
-      
+    const size = this.getCircleSize(tissue?.adj_p_val);
+    const color = this.getCircleColor(tissue?.logfc);
+    
     return {
-      display: size ? 'block' : 'none',
+      display: size > 0 ? 'block' : 'none',
```
sagely1, line 932: Now that size will always be defined as a number, show for values larger than zero.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -1019,10 +1026,10 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
           ...[
             tissueName,
-            tissue ? tissue.logfc : '',
-            tissue ? tissue.ci_r : '',
-            tissue ? tissue.ci_l : '',
-            tissue ? tissue.adj_p_val : '',
+            tissue?.logfc || '',
+            tissue?.ci_r || '',
+            tissue?.ci_l || '',
+            tissue?.adj_p_val || '',
```
hallieswan: Was this change intentional? It looks like this may reintroduce the bug fixed in [AG-1347](https://github.com/Sage-Bionetworks/Agora/pull/1283)

hallieswan (commented):
Thanks for the cleanup! The refactored methods are a lot clearer. Just one small question!

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: Good catch! Not sure why this change didn't merge from AG-1347 but I committed it.

hallieswan (approved):

PR #1286, opened 2024-02-27
AG-1307 - Flatten ensembl_info

sagely1:
Flattening ensembl_info to match data


src/server/models/genes.ts
```diff
@@ -77,7 +77,7 @@ const GeneSchema = new Schema<Gene>(
     median_expression: { type: [MedianExpressionSchema], required: true },
     druggability: { type: [DruggabilitySchema], required: true },
     total_nominations: { type: Number, required: true },
-    ensembl_info: { type: [EnsemblInfoSchema], required: true }
+    ensembl_info: { type: EnsemblInfoSchema, required: true }
```
sagely1, line 80: no longer an array in the data

src/app/models/genes.ts
```diff
@@ -101,8 +101,7 @@ export interface Gene {
 
   bio_domains?: BioDomains;
 
-  // FIXME Will be fixed by AG-1324 in a future data release as it doesn't make sense to have an array here
```
sagely1, line 104: now we can remove the FIXME

src/app/features/genes/components/gene-hero/gene-hero.component.ts
```diff
@@ -103,9 +103,9 @@ export class GeneHeroComponent {
   }
 
   getEnsemblUrl() {
-    if (!this.gene?.ensembl_info || this.gene.ensembl_info.length <= 0)
```
hallieswan, line 106: Could consider adding a test to confirm that an empty string will be returned when `ensembl_info` is falsy (similar to the test for when `ensembl_info` was an empty array, which was understandably removed in this PR), but up to you!

hallieswan (approved):
Looks good to me!

PR #1297, opened 2024-04-02
AG-1412 fixed bug in border condition handling logic

sagely1:
Fixed bug in null handling logic.

### DIFF
--- a/src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.html
+++ b/src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.html
@@ -18,10 +18,8 @@
         <div>P-Value</div>
       </div>
       <div>
-        <div>{{ getSignificantFigures(data.value, 3) }}</div>
-        <div>
-          {{ getSignificantFigures(data.pValue, 3) }}
-        </div>
+        <div [innerHTML]="getSignificantFigures(data.value, 3)"></div>
+        <div [innerHTML]="getSignificantFigures(data.pValue, 3)"></div>
       </div>
     </div>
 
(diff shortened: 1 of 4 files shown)


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -906,8 +906,9 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
     const MIN_SIZE = 6;
     const MAX_SIZE = 50;
 
-    // shouldn't be undefined but if it is, don't show a circle
-    if (pval === undefined)
+    // pval shouldn't be undefined but if it is, don't show a circle
+    // null means there is no data in which case, also don't show a circle
+    if (pval === null || pval === undefined)
```
sagely1, line 911: null handling was missing

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.spec.ts
```diff
@@ -518,5 +519,24 @@ describe('Component: GeneComparisonToolComponent', () => {
       const expected2 = 'ENSG00000147065';
       expect(label2).toBe(expected2);
     });
+
+    it('should set circle size to zero for null/undefined pValues', () => {
+      let tissue: GCTGeneTissue | undefined;
+      // null/undefined values should be zero
+      const result = component.getCircleSize(tissue?.adj_p_val);
+      expect(result).toBe(0);
+    });
```
hallieswan: Since `null` caused this bug, would it make sense to explicitly test the `null` case as well?

```suggestion
    it('should set circle size to zero for undefined pValues', () => {
      let tissue: GCTGeneTissue | undefined;
      // undefined values should be zero
      const result = component.getCircleSize(tissue?.adj_p_val);
      expect(result).toBe(0);
    });
    
    it('should set circle size to zero for null pValues', () => {
      let tissue: GCTGeneTissue = {
        name: "TCX",
        logfc: null,
        adj_p_val: null,
        ci_l: null,
        ci_r: null
      };
      // null values should be zero
      const result = component.getCircleSize(tissue.adj_p_val);
      expect(result).toBe(0);
    });
```

sagely1: You're right, I will break this out into two tests.  Great code suggestion!

src/app/features/genes/components/gene-comparison-tool/components/gene-comparison-tool-details-panel/gene-comparison-tool-details-panel.component.html
```diff
@@ -18,10 +18,8 @@
         <div>P-Value</div>
       </div>
       <div>
-        <div>{{ getSignificantFigures(data.value, 3) }}</div>
-        <div>
-          {{ getSignificantFigures(data.pValue, 3) }}
-        </div>
+        <div [innerHTML]="getSignificantFigures(data.value, 3)"></div>
```
hallieswan: What do you think about changing `getSignificantFigures` to return a character that will be displayed appropriately without using `[innerHTML]`, since that method is used across Agora so we may run into this issue again in another component? I think using the unicode symbol for em dash (`\u2014`) may work, but I haven't tested it!

sagely1: That's a great idea.  I wasn't super happy with having to specify the innerHTML so I think that's a more elegant approach!

hallieswan (approved):
Looks good to me!

PR #1299, opened 2024-04-16
AG-1399 fix invalid ensg in url from hanging

sagely1:
When invalid ENSG were input into the URL, the app would show a loading screen and appeared to hang.

### DIFF
--- a/src/app/features/genes/components/gene-details/gene-details.component.ts
+++ b/src/app/features/genes/components/gene-details/gene-details.component.ts
@@ -135,40 +135,45 @@ export class GeneDetailsComponent implements OnInit, AfterViewInit {
       if (params.get('id')) {
         this.geneService
           .getGene(params.get('id') as string)
-          .subscribe((gene: Gene) => {
-            this.gene = gene;
+          .subscribe((gene) => {
+            if (!gene) {
+              this.helperService.setLoading(false);
+              this.router.navigateByUrl('/404-not-found', { skipLocationChange: true });
+            } else {
+              this.gene = gene;
 
-            this.panels.forEach((p: Panel) => {
-              if (p.name == 'nominations' && !this.gene?.total_nominations) {
-                p.disabled = true;
-              } else if (
-                p.name == 'experimental-validation' &&
-                !this.gene?.experimental_validation?.length
-              ) {
-                p.disabled = true;
-              } else {
-                p.disabled = false;
+              this.panels.forEach((p: Panel) => {
+                if (p.name == 'nominations' && !this.gene?.total_nominations) {
+                  p.disabled = true;
+                } else if (
+                  p.name == 'experimental-validation' &&
+                  !this.gene?.experimental_validation?.length
+                ) {
+                  p.disabled = true;
+                } else {
+                  p.disabled = false;
+                }
+              });
+
+              const nominationsPanel = this.panels.find(
+                (p) => p.name == 'nominations'
+              );
+              if (nominationsPanel) {
+                nominationsPanel.disabled = !this.gene.total_nominations ? true : false;
               }
-            });
 
-            const nominationsPanel = this.panels.find(
-              (p) => p.name == 'nominations'
-            );
-            if (nominationsPanel) {
-              nominationsPanel.disabled = !this.gene.total_nominations ? true : false;
+              const experimentalValidationPanel = this.panels.find(
+                (p) => p.name == 'experimental-validation'
+              );
+              if (experimentalValidationPanel) {
+                experimentalValidationPanel.disabled = !this.gene
+                  .experimental_validation?.length
+                  ? true
+                  : false;
+              }
+
+              this.helperService.setLoading(false);
             }
-
-            const experimentalValidationPanel = this.panels.find(
-              (p) => p.name == 'experimental-validation'
-            );
-            if (experimentalValidationPanel) {
-              experimentalValidationPanel.disabled = !this.gene
-                .experimental_validation?.length
-                ? true
-                : false;
-            }
-
-            this.helperService.setLoading(false);
           });
       }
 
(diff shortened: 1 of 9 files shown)


src/app/features/genes/components/gene-details/gene-details.component.ts
```diff
@@ -135,40 +135,45 @@ export class GeneDetailsComponent implements OnInit, AfterViewInit {
...
-              } else if (
-                p.name == 'experimental-validation' &&
-                !this.gene?.experimental_validation?.length
-              ) {
-                p.disabled = true;
-              } else {
-                p.disabled = false;
+          .subscribe((gene) => {
+            if (!gene) {
+              this.helperService.setLoading(false);
```
sagely1, line 140: added logic to handle when ENSG from URL does not exist

src/app/features/genes/services/gene.service.ts
```diff
@@ -27,13 +27,15 @@ export class GeneService {
...
       return of(this.genes[id]);
     }
-
+    
     return this.apiService.getGene(id).pipe(
-      map((gene: Gene) => {
+      map((gene: Gene | null) => {
+        if (!gene)
+          return null;
         gene.similar_genes_network = this.getSimilarGenesNetwork(gene);
```
sagely1, line 39: This line was crashing in the original code since gene variable was null

tests/gene-comparison-tool.spec.ts
```diff
@@ -27,4 +27,15 @@ test.describe('specific viewport block', () => {
     const dropdown = page.locator('#subCategory');
     await expect(dropdown).toHaveText('Targeted Selected Reaction Monitoring (SRM)');
   });
+
+  test('invalid gene results in a 404 redirect', async ({ page }) => {
+    // go to invalid ENSG page
+    await page.goto('/genes/ENSG00000000000');
+
+    // expect a title "to contain" a substring.
+    await expect(page).toHaveTitle('Agora');
```
sagely1: Not sure if I like having the title just be 'Agora' but that could be something to change down the line.

tests/gene-comparison-tool.spec.ts
```diff
@@ -27,4 +27,15 @@ test.describe('specific viewport block', () => {
...
+
+    // expect div for page not found content to be visible
+    expect(page.locator('.page-not-found')).toBeVisible();
```
sagely1: I did not write a test to check the url as the current behavior is not to change the url when redirecting.  For example navigating to: /genes/ENSG00000 does not change the URL to: /404-not-found.  Therefore, div content on the 404 page is checked to ensure that the 404 page is loaded.

src/app/features/genes/components/gene-details/gene-details.component.ts
```diff
@@ -135,40 +135,45 @@ export class GeneDetailsComponent implements OnInit, AfterViewInit {
...
-              if (p.name == 'nominations' && !this.gene?.total_nominations) {
-                p.disabled = true;
-              } else if (
-                p.name == 'experimental-validation' &&
-                !this.gene?.experimental_validation?.length
-              ) {
-                p.disabled = true;
-              } else {
-                p.disabled = false;
+          .subscribe((gene) => {
```
hallieswan, line 138: Could the type still be specified here?

```suggestion
          .subscribe((gene: Gene | null) => {
```

src/app/features/genes/components/gene-similar/gene-similar.component.ts
```diff
@@ -69,9 +69,11 @@ export class GeneSimilarComponent implements OnInit {
...
         this.geneService
           .getGene(params.get('id') as string)
-          .subscribe((gene: Gene) => {
-            this.gene = gene;
-            this.init();
+          .subscribe((gene: Gene | null) => {
+            if (gene) {
+              this.gene = gene;
+              this.init();
+            }
```
hallieswan: Should this component also handle a gene that doesn't exist? Looks like navigating to https://agora.adknowledgeportal.org/genes/ENSG55555555555/similar causes the same behavior as the route mentioned in the ticket.

tests/gene-comparison-tool.spec.ts
```diff
@@ -27,4 +27,15 @@ test.describe('specific viewport block', () => {
     const dropdown = page.locator('#subCategory');
     await expect(dropdown).toHaveText('Targeted Selected Reaction Monitoring (SRM)');
   });
+
+  test('invalid gene results in a 404 redirect', async ({ page }) => {
```
hallieswan: Does this test belong with the GCT tests or would it make sense to include this test in the `gene-details` test file, since the fix was made in the `gene-details` component?

hallieswan (commented):
Looks good, just a few small questions!

src/app/features/genes/components/gene-details/gene-details.component.ts
sagely1: Great question! In general I'm partial to letting TypeScript inference infer for cases like this since I feel the type is obvious in this case and also since I prefer less verbose code.  That said, I do like to hear why others may choose to type certain variables so let me know if you have strong opinions since code style ultimately helps everyone read each other's code.  I like the explanation in this article: https://dev.to/nectres/dont-type-in-typescript-at-least-most-of-the-time-5b0j#:~:text=You%20don't%20have%20to,the%20benefits%20of%20using%20it.

src/app/features/genes/components/gene-similar/gene-similar.component.ts
sagely1: Good catch! The same fix applies from gene-details.

tests/gene-comparison-tool.spec.ts
sagely1: You're correct.  It should be in its own file.  I will also create the same test for gene-similar.

src/app/features/genes/components/gene-details/gene-details.component.ts
hallieswan: Makes sense! Thanks for sharing the article

hallieswan (approved):
Looks good to me!

PR #1300, opened 2024-04-24
AG-1410: change position so tooltip is displayed in firefox

hallieswan:
https://github.com/Sage-Bionetworks/Agora/assets/26949006/9926f1e0-14cb-48fd-95a6-9a8e94997ee0

### DIFF
--- a/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
+++ b/src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
@@ -402,7 +402,7 @@
                           '). You must unpin some genes before you can pin more.'
                         : 'Add all matching genes to Pinned list'
                     "
-                    tooltipPosition="top"
+                    tooltipPosition="left"
                     tooltipStyleClass="tooltip"
                   >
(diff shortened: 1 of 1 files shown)


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
```diff
@@ -402,7 +402,7 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
                           '). You must unpin some genes before you can pin more.'
                         : 'Add all matching genes to Pinned list'
                     "
-                    tooltipPosition="top"
+                    tooltipPosition="left"
```
hallieswan, line 405: In Firefox, the tooltip would appear on top of the button when the screen was very wide, but not when the screen was narrow. I suspect that the bug was related to how PrimeNG calculates the tooltip position when the tooltip is close to the edge of the screen. Moving the tooltip to the left of the button ensures that the tooltip is shown even when the screen is narrow.

sagely1: Looks great!

sagely1 (approved):

PR #1301, opened 2024-04-24
AG-1407: default to sorting similar genes table by hgnc_symbol ascending

hallieswan:


src/app/features/genes/components/gene-similar/gene-similar.component.html
```diff
@@ -32,6 +32,7 @@ <h1 class="h2">Similar Genes</h1>
           [genes]="genes"
           [columns]="tableColumns"
           sortField="hgnc_symbol"
+          [sortOrder]="1"
```
hallieswan, line 35: gene-table defaults to a descending sort, so specify that this table should default to an ascending sort.

sagely1: looks good!

sagely1 (approved):
Looks great!

PR #1303, opened 2024-04-26
AG-1408: wait for differential expressions boxplot to finish rendering before scrolling to anchor link

hallieswan:
https://github.com/Sage-Bionetworks/Agora/assets/26949006/f5c08d30-991d-4626-90ed-b2c47bde8ce4

### DIFF
--- a/src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
+++ b/src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
@@ -1,10 +1,11 @@
-import { Component, Input, AfterViewInit } from '@angular/core';
+import { AfterViewChecked, Component, Input, ViewChild } from '@angular/core';
 
 import {
   Gene,
   MedianExpression,
   RnaDifferentialExpression,
 } from '../../../../models';
+import { BoxPlotComponent } from '../../../charts/components';
 import { GeneService } from '../../services';
 import { HelperService } from '../../../../core/services';
 
@@ -13,7 +14,7 @@ import { HelperService } from '../../../../core/services';
   templateUrl: './gene-evidence-rna.component.html',
   styleUrls: ['./gene-evidence-rna.component.scss'],
 })
-export class GeneEvidenceRnaComponent implements AfterViewInit {
+export class GeneEvidenceRnaComponent implements AfterViewChecked {
   _gene: Gene | undefined;
   get gene(): Gene | undefined {
     return this._gene;
@@ -35,6 +36,9 @@ export class GeneEvidenceRnaComponent implements AfterViewInit {
 
   consistencyOfChangeChartData: any | undefined;
 
+  @ViewChild(BoxPlotComponent) boxPlotComponent: BoxPlotComponent | null = null;
+  hasScrolled = false;
+
   constructor(
     private helperService: HelperService,
     private geneService: GeneService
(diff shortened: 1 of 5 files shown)


src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
```diff
@@ -71,13 +77,20 @@ export class GeneEvidenceRnaComponent implements AfterViewInit {
     this.initConsistencyOfChange();
   }
 
-  ngAfterViewInit() {
-    const hash = window.location.hash.substr(1);
-    if (hash) {
-      const target = document.getElementById(hash);
-      if (target) {
-        // TODO determine if there are async calls altering the offset height
```
hallieswan, line 79: There is an async call to fetch the differential expressions data, which is then used to render the box plot. We need to wait for the box plot to be initialized before scrolling, otherwise the offset will be incorrect.

tests/gene-comparison-tool.spec.ts
```diff
@@ -8,8 +9,7 @@ test.describe('specific viewport block', () => {
     await page.goto('/genes/comparison?category=Protein+-+Differential+Expression');
 
     // wait for page to load (i.e. spinner to disappear)
-    await expect(page.locator('div:nth-child(4) > div > .spinner'))
-      .not.toBeVisible({ timeout: 250000});
```
hallieswan, line 12: Moved into shared method

src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
```diff
@@ -71,13 +77,20 @@ export class GeneEvidenceRnaComponent implements AfterViewInit {
...
-        window.scrollTo(0, this.helperService.getOffset(target).top - 150);
+  ngAfterViewChecked() {
+    this.scrollToAnchorLink();
+  }
```
hallieswan, line 82: Moved into `ngAfterViewChecked`, since the boxplot hasn't fully loaded when `ngAfterViewInit` is called. Since `ngAfterViewChecked` can be called multiple times, track whether we've already scrolled with `hasScrolled`, so that the page isn't scrolled to the anchor link multiple times.

src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
```diff
@@ -52,6 +56,8 @@ export class GeneEvidenceRnaComponent implements AfterViewInit {
     this.differentialExpressionYAxisMax = undefined;
 
     this.consistencyOfChangeChartData = undefined;
+
+    this.hasScrolled = false;
```
sagely1, line 60: since this variable is already initialized, perhaps this assignment can be removed.

hallieswan: The `reset` method resets the other member variables to their initial value whenever the gene is changed, so I followed the same pattern for the `hasScrolled` variable. However, if `hasScrolled` shouldn't be reset when the gene changes, then I'm happy to remove this assignment!

sagely1: ah, makes sense

src/app/features/genes/components/gene-evidence-rna/gene-evidence-rna.component.ts
sagely1: looks great!

sagely1 (approved):
looks great!

PR #1305, opened 2024-05-07
AG-1425 refactor to support last pinned cache

sagely1:


src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -966,7 +953,7 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
     }
 
     if (this.significanceThresholdActive) {
-      params['significance'] = [this.significanceThreshold];
```
sagely1, line 790: significance value should not be an array, so this will not be converted to an array

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.spec.ts
```diff
@@ -397,7 +398,7 @@ describe('Component: GeneComparisonToolComponent', () => {
       expect(
         element.querySelector(TOGGLE_CLASS)?.querySelector('input')?.checked
       ).toBeTrue();
-      expect(component.getUrlParam('significance')[0]).toEqual(
```
sagely1, line 400: this used to [0] since significance used to be an array.  This has been changed because it does not make sense for significance to be an array.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
```diff
@@ -288,15 +288,27 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
               class="table-divider table-divider-1"
             >
               <div class="table-divider-inner">
-                <div id="pinned-genes-header">Pinned Genes ({{ pinnedGenes.length }}/50)</div>
```
sagely1, line 291: Now, protein and rna views have different headers

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
```diff
@@ -288,15 +288,27 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
...
+                </div>
                 <div class="csv-download-button">
                   <button (click)="downloadPinnedCsv()">
                     <svg-icon name="download" customClass="two-tone"></svg-icon
                     >Download as CSV
                   </button>
                 </div>
                 <div class="clear-all-button">
-                  <button (click)="clearPinnedGenes()">Clear All</button>
+                  <button (click)="onClearAllClick()">Clear All</button>
```
sagely1, line 311: UI triggered methods are now using Click in the name

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -236,18 +240,41 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
-      : this.getUrlParam('pinned', true);
+    
+    const currentPinnedGenesCache = this.pinnedGenesCache;
+
+    let previousPins: (string | undefined)[] = [];
+
+    if (this.currentCategoriesMatchLastPinnedCategories()) {
+      // no need to load from cache if it was last pinned from the current page
+      // previousPins = this.getUrlParam('pinned', true);
+      previousPins = currentPinnedGenesCache.map((g: GCTGene) => g.uid);
```
hallieswan: The pins are currently pulled from the cache, but the comment (and commented-out code) suggests that the pins will be pulled from the URL. Should this be updated to align the comment and the code?

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -236,18 +240,41 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
+    } else {
+      // if it is the initial load, the categories will not match since the last pinned categories 
+      // will be empty strings.  In this case we want to check the url to see if this was a shared url
+      // with pinned genes/proteins.
+      
+      // since the last pinned category is blank, we don't need to check the subcategory
```
hallieswan:
```suggestion
      // if the last pinned category is blank, then this is the initial load and 
      // the last pinned categories are empty strings, so we don't need to check the subcategory. 
      // we should check the url to see if this was a shared url with pinned genes/proteins.
```

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -781,7 +960,7 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
 
     if (this.pinnedGenes.length > 0) {
       params['pinned'] = this.pinnedGenes.map(
-        (g: GCTGene) => g.uid || g.ensembl_gene_id
```
sagely1, line 784: this should be just the uid

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
```diff
@@ -288,15 +288,27 @@ <h1 class="gct-heading h2">Gene Comparison Tool</h1>
...
+                <div id="pinned-genes-header">
+                  <span *ngIf="category === 'RNA - Differential Expression'">
+                    Pinned Genes ({{ pinnedGenes.length }}/50)
+                  </span>
+                  <div *ngIf="category === 'Protein - Differential Expression'">
+                    <div>
+                    Pinned Genes ({{ uniquePinnedGenesCount }}/50)&nbsp;&nbsp;&nbsp;
+                    </div>
+                    <div id="pinned-proteins">
+                      {{ pinnedGenes.length }} Proteins
```
hallieswan: Could `pinnedGenes` be changed to `pinnedItems`, since it can contain the list of proteins or genes?

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -97,8 +97,12 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
 
   /* Pinned ---------------------------------------------------------------- */
+  lastPinnedCategory = '';
+  lastPinnedSubCategory = '';
+
   pinnedGenes: GCTGene[] = [];
-  pinnedGenesCache: { [key: string]: GCTGene[] } = {};
+  uniquePinnedGenesCount = 0;
+  pinnedGenesCache: GCTGene[] = [];
   pendingPinnedGenes: GCTGene[] = [];
```
hallieswan: Consider renaming these variables as well, e.g. `pinnedItemsCache` and `pendingPinnedItems`? @sagely1 also notes that `pendingPinnedGenes` may be extraneous and could be deleted?

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -236,18 +240,41 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
     const pinnedGenes: GCTGene[] = [];
-    const currentPinnedGenesCache = this.getPinnedGenesCache(
-      this.category,
-      this.subCategory
-    );
-    const urlPins = currentPinnedGenesCache.length
-      ? currentPinnedGenesCache.map((g: GCTGene) => g.uid)
-      : this.getUrlParam('pinned', true);
+    
+    const currentPinnedGenesCache = this.pinnedGenesCache;
```
hallieswan: Consider renaming here as well -- `pinnedGenes` -> `pinnedItems`, `currentPinnedGenesCache` -> `currentPinnedItemsCache`

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
hallieswan:
```suggestion
      previousPins = currentPinnedGenesCache.map((g: GCTGene) => g.uid);
```

hallieswan: Removing the comment, since the if statement makes it pretty clear!

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -236,18 +240,41 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
+      // since the last pinned category is blank, we don't need to check the subcategory
+      if (this.lastPinnedCategory === '') {
+        // check the url for pinned genes/proteins
+        previousPins = this.getUrlParam('pinned', true);
+        this.setLastPinnedCategories();
+      } else {
+        // categories don't match, so grab it from the cache
+        if (this.category === 'RNA - Differential Expression') {
+          previousPins = currentPinnedGenesCache
+            .map((g: GCTGene) => g.uid?.substring(0, 15));
```
hallieswan: Could just get the ensembl gene id directly instead? 

```suggestion
          previousPins = currentPinnedGenesCache
            .map((g: GCTGene) => g.ensembl_gene_id;
```

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -256,27 +283,30 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
         gene.hgnc_symbol.toLowerCase(),
       ];
 
-      if ('Protein - Differential Expression' === this.category) {
+      if (this.category === 'Protein - Differential Expression') {
+        // for this category, the uid will be ensg + uniprotid
         gene.uid += gene.uniprotid;
```
hallieswan: Consider pulling this into a new helper method that returns the uid based on the category

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -236,18 +240,41 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
+          previousPins = currentPinnedGenesCache
+            .map((g: GCTGene) => g.uid?.substring(0, 15));
+        } else {
+          previousPins = currentPinnedGenesCache.map((g: GCTGene) => g.uid);
+        }
+      }
+    }
 
     genes.forEach((gene: GCTGene) => {
       gene.uid = gene.ensembl_gene_id;
```
hallieswan: Consider using helper method for uid assignment here -- 

```ts
const helper = (category: string, gene: GCTGene) => {
  switch(category) {
    case 'RNA - Differential Expression': 
      return gene.ensembl_gene_id;
    case 'Protein - Differential Expression':
     return gene.ensembl_gene_id + gene.uniprot_id;
  }
}
```

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.html
hallieswan: @sagely1 also noticed that when `pinnedGenes.length` is 1, the text currently reads `1 Proteins`, but should read `1 Protein`

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -236,18 +240,41 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
-      this.category,
-      this.subCategory
-    );
-    const urlPins = currentPinnedGenesCache.length
-      ? currentPinnedGenesCache.map((g: GCTGene) => g.uid)
-      : this.getUrlParam('pinned', true);
+    
+    const currentPinnedGenesCache = this.pinnedGenesCache;
+
+    let previousPins: (string | undefined)[] = [];
```
hallieswan: Add a comment here noting the previousPins contains the uids of the previously pinned genes, formatted based on the _current_ category. Maybe provide an example? e.g. previous category was Protein, which has uids of ensembl + uniprot, but previousPins should contain just ensembl

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: correct, there is no need to keep this legacy code.

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: I will create a helper method for this purpose

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: created a comment along these lines in a new helper method

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -613,138 +644,276 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
 
-  getPinnedGenesCache(category: string, subCategory?: string) {
-    const key = this.getPinnedGenesCacheKey(category, subCategory);
-    return this.pinnedGenesCache[key] || [];
-  }
-
-  setPinnedGenesCache(genes: GCTGene[], category: string, subCategory: string) {
-    const key = this.getPinnedGenesCacheKey(category, subCategory);
-    this.pinnedGenesCache[key] = genes;
+  setPinnedGenesCache(genes: GCTGene[]) {
```
hallieswan: Could consider updating this variable name as well -- `setPinnedItemsCache`

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -613,138 +644,276 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
-        delete this.pinnedGenesCache[key];
-      } else {
-        for (const k in this.pinnedGenesCache) {
-          if (k.indexOf(key) === 0) {
-            delete this.pinnedGenesCache[k];
-          }
-        }
-      }
-    }
+  clearPinnedGenesCache() {
```
hallieswan: And here as well -- `clearPinnedItemsCache()`

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
```diff
@@ -613,138 +644,276 @@ export class GeneComparisonToolComponent implements OnInit, AVI, OnDestroy {
...
   getPinnedUniProtIds() {
-    return this.pinnedGenes.map((g: GCTGene) => g.uniprotid);
+    return this.pinnedItems.map((g: GCTGene) => g.uniprotid);
+  }
+
+  getPinDisabledStatus() {
+    if (this.category === 'RNA - Differential Expression')
+      return this.pinnedItems.length >= this.maxPinnedGenes;
+    else {
+      // default to showing pin all button for protein view
```
hallieswan: I think`getPinDisabledStatus` is used for both individual item pins and pin all button, so could the comment be generalized? 

```suggestion
      // default to showing pin buttons in protein view
```

hallieswan (approved):

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: done!

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: done!

src/app/features/genes/components/gene-comparison-tool/gene-comparison-tool.component.ts
sagely1: done!

PR #1311, opened 2024-06-05
AG-1471, AG-1473, AG-1474: add workflow to create agora/data Docker image and run end-to-end tests

hallieswan:
Currently, the workflow builds the agora/data image on the runner and then runs the e2e tests. In a future PR, we'll split this workflow into three jobs: 
  1. check if the agora/image already exists in a remote registry
  2. build and push the image to the remote registry if it does not already exist
  3. pull the image from the remote registry, then run e2e tests

Note: we will need to add the appropriate environment variables and secrets to this repository before the workflow will run.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,74 @@
+on:
+  push:
+    branches:
+      - AG-1471_AG-1473_AG-1474
+
+jobs:
+  e2e:
+    runs-on: ubuntu-20.04
+    environment: e2e
+    env:
+      CI: true
+      LOCAL_DATA_PATH: './local/data'
+      LOCAL_TEAM_IMAGES_DIR: './local/data/team_images'
+      LOCAL_COLLECTIONS_PATH: './scripts/collections.csv'
+      LOCAL_IMPORT_SCRIPT_PATH: './docker/import-data.sh'
+      LOCAL_CREATE_INDEXES_PATH: './scripts/mongo-create-Indexes.js'
+      DOCKER_COMPOSE_PATH: './docker/docker-compose.yml'
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: get data image tag, data manifest id, and data version
+        env:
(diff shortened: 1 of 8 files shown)


playwright.config.ts
```diff
@@ -75,7 +76,8 @@ export default defineConfig({
   /* Run your local dev server before starting the tests */
   webServer: {
     command: 'npm run start',
-    url: baseURL,
-    reuseExistingServer: !process.env.CI,
+    port,
+    reuseExistingServer: true,
```
hallieswan, line 80: Reuse the existing server rather than having playwright start the server for us

src/server/api.ts
```diff
@@ -38,30 +38,37 @@ const database = { url: '' };
...
-  const results = awsParamStore.getParametersSync(
-    [
-      '/agora-' + process.env.APP_ENV + '/MongodbUsername',
-      '/agora-' + process.env.APP_ENV + '/MongodbPassword',
-    ],
-    { region: 'us-east-1' }
-  );
+  let dbUser: string | undefined;
+  let dbPass: string | undefined;
+  if (process.env.APP_ENV === 'e2e') {
```
hallieswan, line 49: Get credentials from environment variables when running e2e tests

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,146 @@
+name: Build Data Image and Run End-to-end Tests
+
+on: push
+
+jobs:
+  e2e:
+    # Run in Sage repo on develop or main branches
+    # and on all branches in user-owned forks
+    if: ${{ github.ref_name == 'develop' || github.ref_name == 'main' || github.actor == github.repository_owner }}
```
hallieswan, line 9: Forks do not have access to the Sage repo's secrets, so don't run this workflow in the Sage repo on pull_request trigger. Instead, run the workflow when there is a push to develop/main or when the user pushes to a repo that they own (i.e. their forked repo).

sagely1:
Looks great!

sagely1 (approved):

PR #1313, opened 2024-06-14
AG-1472: push/pull data images from GHCR

hallieswan:
Will create a GHCR package called "agora-data" in the user's account (for forked repos) or in the Sage-Bionetworks organization (for the base repo). See example GHCR package in user account [here](https://github.com/hallieswan/Agora/pkgs/container/agora-data).

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,9 +1,6 @@
 name: Build Data Image and Run End-to-end Tests
 
 on: push
-env:
-  ORG_NAME: 'Sage-Bionetworks'
-  NAMESPACE: ${{ github.repository_owner }}
 
 jobs:
   check-for-image:
@@ -16,6 +13,8 @@ jobs:
       packages: read
     environment: e2e
     env:
+      ORG_NAME: 'Sage-Bionetworks'
+      NAMESPACE: ${{ github.repository_owner }}
       DATA_IMAGE_NAME: '${{ vars.DB_NAME }}-data'
     outputs:
       DATA_MANIFEST_ID: ${{ steps.get-ids-from-package-json.outputs.DATA_MANIFEST_ID }}
@@ -61,7 +60,7 @@ jobs:
         run: |
           GH_PATH=$( [[ "${{ env.NAMESPACE }}" == "${{ env.ORG_NAME }}" ]] && echo "${{ env.ORG_PATH }}" || echo "${{ env.USER_PATH }}" )
 
-          # allow errror to handle case where gh throws an error when package has not yet been created
+          # allow error to handle case where gh throws an error when package has not yet been created
           set +e
           IMAGE_VERSIONS=$(gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" "${GH_PATH}")
           set -e
@@ -135,7 +134,7 @@ jobs:
           docker build \
           -t "${{ env.DATA_IMAGE_PATH }}" \
           -f "${{ env.DOCKERFILE_PATH }}" \
-          --build-arg SOURCE_REPO="https://github.com/${{ env.NAMESPACE }}/${{ vars.DB_NAME }}" \
+          --build-arg SOURCE_REPO="https://github.com/${{ github.repository }}" \
           --build-arg DESCRIPTION="${GITHUB_SHA::7}" \
           --build-arg LOCAL_DATA_PATH="${{ env.LOCAL_DATA_PATH }}" \
           --build-arg LOCAL_COLLECTIONS_PATH="${{ env.LOCAL_COLLECTIONS_PATH }}" \
@@ -151,7 +150,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       # Note: if a new package is created using this workflow, then the package
-      # visibility will default to the source repository visbility. For example,
+      # visibility will default to the source repository visibility. For example,
       # a package created by a workflow in a public repo will also be public.
       # Publicly visible container packages are anonymously downloadable.
       - name: push to GitHub Container Registry
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,9 +9,14 @@ jobs:
     if: ${{ github.ref_name == 'develop' || github.ref_name == 'main' || github.actor == github.repository_owner }}
     timeout-minutes: 60
     runs-on: ubuntu-20.04
+    permissions:
+      packages: write
     environment: e2e
     env:
       CI: true
+      ORG_NAME: 'Sage-Bionetworks'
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -1,48 +1,100 @@
...
-      LOCAL_COLLECTIONS_PATH: './scripts/collections.csv'
-      LOCAL_IMPORT_SCRIPT_PATH: './docker/import-data.sh'
-      LOCAL_CREATE_INDEXES_PATH: './scripts/mongo-create-Indexes.js'
-      DOCKER_COMPOSE_PATH: './docker/docker-compose.yml'
+      DATA_IMAGE_NAME: '${{ vars.DB_NAME }}-data'
+    outputs:
+      DATA_MANIFEST_ID: ${{ steps.get-ids-from-package-json.outputs.DATA_MANIFEST_ID }}
+      DATA_VERSION: ${{ steps.get-ids-from-package-json.outputs.DATA_VERSION }}
+      DATA_IMAGE_PATH: ${{ steps.get-ids-from-package-json.outputs.DATA_IMAGE_PATH }}
+      IMAGE_EXISTS: ${{ steps.check-container-repo.outputs.IMAGE_EXISTS }}
```
hallieswan, line 23: Outputs will be used by subsequent jobs

.github/workflows/e2e.yml
```diff
@@ -1,48 +1,100 @@
...
+          # https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#list-package-versions-for-a-package-owned-by-an-organization
+          ORG_PATH: '/orgs/${{ env.ORG_NAME }}/packages/container/${{ env.DATA_IMAGE_NAME }}/versions'
+          # https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#list-package-versions-for-a-package-owned-by-the-authenticated-user
+          USER_PATH: '/user/packages/container/${{ env.DATA_IMAGE_NAME }}/versions'
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DATA_IMAGE_TAG: ${{ steps.get-ids-from-package-json.outputs.DATA_IMAGE_TAG }}
         run: |
-          echo "==> data image tag: ${{ env.DATA_IMAGE_TAG }}"
-          echo "IMAGE_EXISTS=false" >> "$GITHUB_ENV"
+          GH_PATH=$( [[ "${{ env.NAMESPACE }}" == "${{ env.ORG_NAME }}" ]] && echo "${{ env.ORG_PATH }}" || echo "${{ env.USER_PATH }}" )
```
hallieswan, line 61: If the job is running in the Sage-Bionetworks repo, then check for the image in the organization namespace using the ORG_PATH. Otherwise, check for the image in the user's namespace using the USER_PATH.

.github/workflows/e2e.yml
```diff
@@ -1,48 +1,100 @@
...
+            IMAGE_EXISTS="false"
+          else
+            IMAGE_EXISTS=$(echo "${IMAGE_VERSIONS}" | jq '[.[] | select(any(.metadata.container.tags[]; . == "${{ env.DATA_IMAGE_TAG }}"))] | length > 0')
+          fi
+
+          echo "IMAGE_EXISTS=${IMAGE_EXISTS}" >> "$GITHUB_OUTPUT"
+
+  build-image:
+    needs: check-for-image
+    if: ${{ needs.check-for-image.result == 'success' && needs.check-for-image.outputs.IMAGE_EXISTS == 'false' }}
```
hallieswan, line 79: The workflow will build the image if it does not already exist in the GHCR package, as in [this run](https://github.com/hallieswan/Agora/actions/runs/9508204995).

.github/workflows/e2e.yml
```diff
@@ -81,16 +133,53 @@ jobs:
...
+      - name: push to GitHub Container Registry
+        run: docker push "${{ env.DATA_IMAGE_PATH }}"
+
+  run-e2e-tests:
+    needs:
+      - check-for-image
+      - build-image
+    # always() prevents skipping this job when build-image was skipped
+    # see https://github.com/actions/runner/issues/491
+    if: ${{ always() && needs.check-for-image.result == 'success' && (needs.build-image.result == 'success' || needs.build-image.result == 'skipped')}}
```
hallieswan, line 165: The workflow will run the run-e2e-test job even if the image did not need to be built, as in [this run](https://github.com/hallieswan/Agora/actions/runs/9508204995/attempts/1).

docker/Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM mongo:latest
 
+# GHCR labels
+ARG SOURCE_REPO="https://github.com/Sage-Bionetworks/agora"
```
sagely1, line 4: to make it more reusable for other repos consider using:
`ARG SOURCE_REPO="https://github.com/${{ github.repository }}"`

.github/workflows/e2e.yml
```diff
@@ -77,18 +102,28 @@ jobs:
...
+          --build-arg DESCRIPTION="${GITHUB_SHA::7}" \
           --build-arg LOCAL_DATA_PATH="${{ env.LOCAL_DATA_PATH }}" \
           --build-arg LOCAL_COLLECTIONS_PATH="${{ env.LOCAL_COLLECTIONS_PATH }}" \
           --build-arg LOCAL_IMPORT_SCRIPT_PATH="${{ env.LOCAL_IMPORT_SCRIPT_PATH }}" \
           --build-arg LOCAL_CREATE_INDEXES_PATH="${{ env.LOCAL_CREATE_INDEXES_PATH }}" \
           .
-      # TODO - push image to remote repository
+
+      # Note: if a new package is created using this workflow, then the package
+      # visibility will default to the source repository visbility. For example,
```
sagely1: typo:visibility

sagely1 (approved):
Looks great! I left some optional feedback.

docker/Dockerfile
hallieswan: Agreed on reusability! I've tried to do that by making `SOURCE_REPO` a build-time [ARG](https://docs.docker.com/reference/dockerfile/#arg), so it can be overwritten when the Docker image is built. The current value is a default that primarily demonstrates what `SOURCE_REPO` should look like. 

In the GHA workflow, we override the value at build-time [here](https://github.com/Sage-Bionetworks/Agora/blob/48b4d5a33892756af126c13e8838486cf9c9a87d/.github/workflows/e2e.yml#L138), but I agree that using `${{ github.repository }}` would be clearer, so I'll update to use that instead!

.github/workflows/e2e.yml
hallieswan: Thanks for the catch! I'll update in the next commit.

sagely1 (approved):
Looks better than it already did!

PR #1315, opened 2024-06-14
AG-1472: ensure org package is lowercase

hallieswan:
Fix for failed workflow run [here](https://github.com/Sage-Bionetworks/Agora/actions/runs/9521629725/job/26249486143).


sagely1 (approved):
Looks good! Occurs to me we might want to create some utility methods we can reuse to handle these common string manipulations in the future.

PR #1320, opened 2024-07-01
AG-1498: improvements from validation

hallieswan:


.github/workflows/ci.yml
```diff
@@ -215,7 +215,7 @@ jobs:
       - name: start database, server, and app
         run: |
           echo "==> start database and data containers"
-          docker-compose -f "${{ env.DOCKER_COMPOSE_PATH }}" up -d
+          docker compose -f "${{ env.DOCKER_COMPOSE_PATH }}" up -d
```
hallieswan, line 218: `docker-compose` refers to an older, standalone tool (Compose v1), whereas `docker compose` refers to the new, integrated version built directly into Docker (Compose v2), as described [here](https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2). Newer Docker Desktop versions [alias](https://docs.docker.com/compose/migrate/#what-does-this-mean-for-my-projects-that-use-compose-v1) `docker-compose` to redirect commands to `docker compose`. However, users with older Docker Desktop versions may run into issues when v1 is used instead of v2. Change scripts to use `docker compose` instead of `docker-compose` so Compose v2 will be used.

README.md
```diff
@@ -185,14 +186,15 @@ The `sage-bionetworks` package will contain images for data releases that have b
...
 
-1. Create a new branch.
-2. Update the `package.json` to reflect the appropriate `data-file` and `data-version` files.
-3. If necessary, update `./scripts/collections.csv` to specify new collections and `./scripts/mongo-create-Indexes.js` to specify new indexes.
-4. Commit the changes.
-5. Push the changes to your remote fork to trigger a run of the `ci.yml` workflow. 
-6. The new image will be available in your user namespaced GHCR package, e.g. `https://github.com/hallieswan/Agora/pkgs/container/agora-data-nonmonorepo`.
-7. Update your local `.env` file so `DATA_IMAGE_PATH` points to the newly created image, e.g. `ghcr.io/hallieswan/agora-data-nonmonorepo:syn13363290.68`.
-8. Start the containerized database: `npm run docker:db:start`.
+1. Set up GitHub secrets/variables in your remote fork as described [here](docker/README.md#workflow-setup), if necessary.
```
hallieswan, line 189: Specify that user must add creds to their GitHub fork before creating the new image. Subsequent steps are unchanged.

package.json
```diff
@@ -41,11 +41,11 @@
     "e2e": "playwright test --trace on",
     "e2e:ui": "playwright test --ui",
     "e2e:update": "npm install -D @playwright/test@latest; npx playwright install --with-deps; npx playwright --version",
-    "create-env": "DB_PASS=$(openssl rand -base64 12) DATA_IMAGE_TAG=$(npm run docker:data-image-tag -s) envsubst < .env.example > .env",
+    "create-env": "[ ! -f .env ] && DB_PASS=$(openssl rand -base64 12) DATA_IMAGE_TAG=$(npm run docker:data-image-tag -s) envsubst < .env.example > .env",
```
hallieswan, line 44: Only create the `.env` file if one does not already exist. This allows the dev to configure the connection to the containerized database via a GUI tool once, rather than requiring them to update the creds each time `create-env` is run.

.github/workflows/ci.yml
hallieswan: Updated workflow still runs as expected -- https://github.com/hallieswan/Agora/actions/runs/9749882114/job/26908035251

sagely1 (approved):

hallieswan:
This repo is deprecated, so these changes are no longer needed.

PR #1334, opened 2024-11-27
AG-1488: Added new resource cards in Resources tab

sagely1:
## Description

Adding new resource cards

image

image

### DIFF
--- a/src/app/features/genes/components/gene-resources/gene-resources.component.html
+++ b/src/app/features/genes/components/gene-resources/gene-resources.component.html
@@ -1,26 +1,25 @@
 <div *ngIf="gene" class="section">
-  <div class="section-inner pb-0">
+  <div class="section-inner">
     <div class="container-sm">
-      <div>
-        <h2>
-          Druggability of {{ gene.hgnc_symbol || gene.ensembl_gene_id }}
-        </h2>
-        <p class="druggability-description">
-          AMP-AD pharmaceutical partners scored the tractability of targets
-          using publicly-available resources and have generated ratings of
-          small-molecule druggability, therapeutic antibody feasibility, and
-          safety.
-        </p>
-        <modal-link
-          text="Learn more about the analysis"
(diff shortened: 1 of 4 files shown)


sagely1:
@JessterB: one of the existing cards pointing at Pub AD does not like the hgnc symbol.  Not sure if this is expected on this link but I figure you might want to validate it.  https://adexplorer.medicine.iu.edu/pubad/external/AARD

sagely1:
@JessterB: also this link to UniProtKB seemed like it didn't like ENSG.  https://www.uniprot.org/uniprotkb?query=ENSG00000205002

JessterB:
> @JessterB: one of the existing cards pointing at Pub AD does not like the hgnc symbol. Not sure if this is expected on this link but I figure you might want to validate it. https://adexplorer.medicine.iu.edu/pubad/external/AARD

PubAD doesn't seem to have every gene that we have, but since there's not much we can do about that we will live with some gene links going to the 'no results found' page. Note that PubAD's ENSG search results are completely broken so we can't use ENSG links either:

Screenshot 2024-11-26 at 4 52 42 PM


> @JessterB: also this link to UniProtKB seemed like it didn't like ENSG. https://www.uniprot.org/uniprotkb?query=ENSG00000205002

That link looks fine to me? It returns a single search result for ENSG00000205002, [Q4LEZ3](https://www.uniprot.org/uniprotkb/Q4LEZ3/entry).

tschaffter:
About the PubAD screenshot shared by Jess: Is the blue dot a radio button? If yes, we should add some horizontal space between the button and its label (Gene Query)

tschaffter (approved):

JessterB:
> About the PubAD screenshot shared by Jess: Is the blue dot a radio button? If yes, we should add some horizontal space between the button and its label (Gene Query)

@tschaffter While I agree that PubAD could use some design love, it's not our application. We just link to it...

PR #1335, opened 2024-11-27
AG-1257: Add pharos class to hero

sagely1:
## Description
Added pharos class to hero

## Testing
Test one gene with pharos class and one without and ensure PHAROS CLASS shows / hides as needed:
http://localhost:8080/genes/ENSG00000000003

image

http://localhost:8080/genes/ENSG00000002079

image

### DIFF
--- a/src/app/features/genes/components/gene-hero/gene-hero.component.html
+++ b/src/app/features/genes/components/gene-hero/gene-hero.component.html
@@ -23,7 +23,7 @@
           </p>
         </div>
         <div *ngIf="gene.druggability[0].pharos_class">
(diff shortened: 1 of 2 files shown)


tschaffter (approved):

JessterB:
@sagely1 My only comment is that the line spacing underneath the new Pharos Class element is too small, it should be the same as the spacing underneath the biodomain element. 

Biological Domains
Apoptosis
_existing line spacing sets off this section from the next section_
Pharos Class
Tbio
_we want the same line spacing here to visually separate this element from the AKA section_
Also known as

PR #1338, opened 2024-12-03
AG-1569 - fix GHA workflow failing

sagely1:
`docker-compose` syntax is now deprecated and why GHA workflow is failing.
`docker compose` is the new syntax

This was a successful run on my branch:
https://github.com/sagely1/Agora/actions/runs/12149499015


JessterB (approved):
I'm glad this was such an easy fix!

PR #1341, opened 2024-12-17
AG-1599 - handle a list of pharos class values

sagely1:
## Changes
- Druggability prop changed from array to object in data version 71
- Pharos class values are now comma deliminated and conditionally displayed
- "No value" logic removed

## Testing
Same method as in AG-1257
Test one gene with pharos class and one without and ensure PHAROS CLASS shows / hides as needed:
http://localhost:8080/genes/ENSG00000000003

image

http://localhost:8080/genes/ENSG00000002079

image

### DIFF
--- a/src/app/features/genes/components/gene-druggability/gene-druggability.component.ts
+++ b/src/app/features/genes/components/gene-druggability/gene-druggability.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, Input } from '@angular/core';
+import { Component, Input } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 
 import { Gene, Druggability } from '../../../../models';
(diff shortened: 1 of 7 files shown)


src/app/features/genes/components/gene-druggability/gene-druggability.component.ts
```diff
@@ -43,28 +43,23 @@ export class GeneDruggabilityComponent implements OnInit {
...
+        this.gene.druggability =
+        {
+          sm_druggability_bucket: this.getDefaultBucketNumber(),
+          safety_bucket: this.getDefaultBucketNumber(),
+          abability_bucket: this.getDefaultBucketNumber(),
+          pharos_class: [],
+          classification: this.getDefaultText(),
+          safety_bucket_definition: this.getDefaultText(),
+          abability_bucket_definition: this.getDefaultText(),
+        };
```
JessterB, line 59: Leaving the removed fields here seems harmless since they just get the defaults and we don't display those anywhere, but it woiuld be nice if they were gone entirely in the monorepo version of the app.

JessterB (approved):
The pharos class handling looks great, I only had one thought about the monorepo version of this change.

